PR #199, opened 2013-11-18
Add single sign-on (SSO) support

olivierdagenais:
Jenkins and Artifactory both support SSO through the use of an HTTP header sent by a reverse proxy.  In Jenkins, this is handled by the [Reverse Proxy Auth Plugin](https://wiki.jenkins-ci.org/display/JENKINS/Reverse+Proxy+Auth+Plugin) while in Artifactory, it's part of the [Pro Power Pack](http://www.jfrog.com/home/v_artifactorypro_overview).

**How to configure SSO for GitBucket**
1. Configure a front-end reverse proxy to authenticate a request (such as IIS with Windows Authentication) and then delegate the request to the servlet container, augmenting it with a special header (let's call it `auth-user`).
2. Configure the servlet app to look for the [configurable] `auth-user` header, and if it's set, you can assume that the user named by the header is logged in, otherwise you continue with other authentication mechanisms. (this is the part I am submitting for review)

**Implementation details**
1. The LDAP configuration needs to record the name of the reverse-proxied SSO header (this is totally optional, if no value is provided, SSO won't be attempted and it will work as before).
2. When retrieving the `LoginAccount`, if the session is empty AND the SSO header has a value, then construct an `Option[Account]`, looking up the user's e-mail address, save the account in the session and return it, otherwise proceed as before.

**Manual testing**
1. Install Tomcat 7.0 and have it listen for requests on port 8080.
2. Deploy GitBucket in Tomcat.
3. Test GitBucket by visiting http://localhost:8080/gitbucket/
3.1 Notice that you can sign in as root and operate as usual.
3.2 Configure LDAP but don't set the `HTTP Single Sign-On header` field.
3.3 Notice that you can sign on as domain users, etc.
4. Configure IIS on port 80 to use Windows authentication.
5. Install the [BonCode Apache Tomcat AJP 1.3 Connector](http://boncode.net/connector/webdocs/Tomcat_Connector.htm) and configure it to bridge IIS to Tomcat.
5.1 You may want to test that you can navigate to http://localhost/gitbucket/ and it works like it did on Tomcat.
6. Configure the `HTTP Single Sign-On header` field to `auth-user`.
7. Test GitBucket w/SSO by visiting http://localhost/gitbucket/
7.1 You will be signed in automatically.
7.2 Clicking "Sign out" does nothing. (perhaps we would eventually want to hide that link when we detect SSO is enabled?)

### DIFF
--- a/src/main/scala/app/ControllerBase.scala
+++ b/src/main/scala/app/ControllerBase.scala
@@ -3,13 +3,14 @@ package app
 import _root_.util.Directory._
 import _root_.util.Implicits._
 import _root_.util.ControlUtil._
-import _root_.util.{StringUtil, FileUtil, Validations, Keys}
+import _root_.util.{StringUtil, FileUtil, LDAPUtil, Validations, Keys}
 import org.scalatra._
 import org.scalatra.json._
 import org.json4s._
 import jp.sf.amateras.scalatra.forms._
 import org.apache.commons.io.FileUtils
 import model.Account
+import scala.Some
 import service.{SystemSettingsService, AccountService}
 import javax.servlet.http.{HttpServletResponse, HttpSession, HttpServletRequest}
 import java.text.SimpleDateFormat
@@ -21,7 +22,7 @@ import org.scalatra.i18n._
  */
 abstract class ControllerBase extends ScalatraFilter
   with ClientSideValidationFormSupport with JacksonJsonSupport with I18nSupport with FlashMapSupport with Validations
-  with SystemSettingsService {
+  with SystemSettingsService with AccountService {
 
   implicit val jsonFormats = DefaultFormats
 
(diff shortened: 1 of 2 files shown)


eidorb:
I'm guessing this is an issue that SSO poses to applications in general, but how would one go about *not* using SSO temporarily? For example, an administrator may wish to use a normal account for their day-to-day repos, but a separate root account to actually administer the site.

triplem:
Do not like to necro bump, gut SSO is a rather important feature. 

Any progress apart from this PR?

src/main/scala/app/ControllerBase.scala
```diff
@@ -34,8 +35,35 @@ abstract class ControllerBase extends ScalatraFilter
     val context      = request.getServletContext.getContextPath
     val path         = httpRequest.getRequestURI.substring(context.length)
 
+    var account = httpRequest.getSession.getAttribute(Keys.Session.LoginAccount).asInstanceOf[Account]
+    if(account == null) {
```
ritschwumm: please avoid calling isDefined/get on Option, this can be done in a safer way using foreach or a for-comprehension

bviktor:
Can you please clarify how this works? How are the credentials passed from the client to the server? Your example mentions localhost, which is a different story, but how does one authenticate remotely by just opening the GitBucket home page? Or do you have to enter the credentials to the reverse proxy? Because that's what the mentioned Jenkins plugin does: merely delegate the login prompt to the reverse proxy, which isn't making things simpler from a user's perspective, only _different_.

triplem:
With a Remote Proxy, the Usename is passend via Http Headers to the Server. This makes it necessary, that the Proxy is fully trusted by the Application. 
There are quite a lot of different SSO Implementations Out there, but the Reverse Proxy seems to be the easiest to implement.

bviktor:
> With a Remote Proxy, the Usename is passend via Http Headers to the Server.

Yeah, the question is: how. Because browsers surely don't send the running user's username and password to visited websites on their own, so you won't be signed in "automatically" as the poster says.

Also, I think there's a misunderstanding here: by "remote" I mean remote to the user. The proxy and GitBucket can be (and I think should be) on the same host. This is a critical point coz I think the poster's example only worked because the OS authed the user - both the browser and IIS were running on the same host. That won't work when the proxy is on a different host.

triplem:
Well, the SSO Server (eg Keycloak) Containern the credentials and the Proxy communities to this Server and Sets the Usename in the Header and/or a Cookie. For an example see http://bandrzejczak.com/blog/2015/12/06/sso-for-your-single-page-application-part-2-slash-2-akka-http/.

bviktor:
Thanks for the info, triplem, I'll look into that.

olivierdagenais:
> Yeah, the question is: how. Because browsers surely don't send the running user's username and password to visited websites on their own, so you won't be signed in "automatically" as the poster says.

There are some browsers that will respond to a challenge from the server (usually using a protocol like NTLM or Kerberos), therefore proving to the server that they are indeed the user they claim to be, without having to send the password to the server.

The protocol usually goes something like the following, with _Authority_ usually being the logon server, such as an Active Directory Domain Controller:

1. Client -> Server: I am _olivierdagenais_
2. Server -> Client: Prove it!  Transform this totally random number using your password and send me the result.
3. Client -> Server: Challenge accepted!  Here's the result.
4. Server -> Authority: I sent _olivierdagenais_ this totally random number as a challenge and he replied with this result.  Does that match?
5. Authority->Server: Yep, that **must** be _olivierdagenais_, so you're good to go.
6. Server -> Client: OK, I believe you.  Welcome, _olivierdagenais_.

I simplified a few parts, but that's usually how it goes.  The server (or anybody else who might be listening to the conversation) can't determine the password based on the challenge and/or result.

In my case, I had configured IIS to authenticate Active Directory users on the domain such that IIS could then forward/proxy the request to GitBucket with a special header indicating what their username was.

If there's sufficient interest, I _could_ be convinced to rebase this branch. ;)

Issue #123, opened 2013-10-05
The change password page needs to require old password and 2 copies of the new password.


davewichers:
When you go to the page: http://localhost:8080/gitbucket/root/_edit

It just says enter the new password. If the user's session is ever hijacked or they walk away from their computer, an attacker could change their password easily.

To make sure they can't, require the user to enter their old password first. And then verify this is the correct original password before trying to change it.

In addition, to make sure the user actually knows their new password, require them to enter the new password twice and make sure the 2 values match before changing the password.

PR #225, opened 2013-12-02
Fix for #123: The change password page needs to require old password and 2 copies of the new password.

mcveat:


aadrian:
As mentioned in https://github.com/gitbucket/gitbucket/issues/123  , it's important to ask the old password before allowing to change it to a new one.

@mcveat could you please fix the merge conflicts?

PR #301, opened 2014-03-04
Support multiple database driver

okapies:
### Objective
Make database driver configurable, other than default [H2Driver](http://slick.typesafe.com/doc/2.0.0/api/scala/slick/driver/H2Driver.html), to support other databases such as [MySQL](http://slick.typesafe.com/doc/2.0.0/api/scala/slick/driver/MySQLDriver.html) and [PostgreSQL](http://slick.typesafe.com/doc/2.0.0/api/scala/slick/driver/PostgresDriver.html) in the future.

### Design
Note: I have implemented it with reference to [slick-examples](https://github.com/slick/slick-examples/blob/master/src/main/scala/com/typesafe/slick/examples/lifted/MultiDBCakeExample.scala).

Add `Profile` trait to combine it with `Table`.
```
trait Profile {
  val profile: ExtendedProfile
}
```

Import `Profile.profile.simple._` instead of `scala.slick.driver.H2Driver.simple._` directly in the table definition:
```
trait AccountComponent { self: Profile =>
  import profile.simple._
  object Accounts extends Table[Account]("ACCOUNT") {
    def userName = column[String]("USER_NAME", O PrimaryKey)
    ...
```

Services include corresponding `Component` to import the `Table` object.
```
trait AccountService extends AccountComponent { self: Profile =>
  def getAccountByUserName(userName: String, includeRemoved: Boolean = false): Option[Account] =
    Query(Accounts) filter(t => (t.userName is userName.bind) && (t.removed is false.bind, !includeRemoved)) firstOption
  ...
}
```

Now almost every controller requires an `ExtendedProfile` instance to inject into the services, combined with `Profile`.
```
class AccountController(override val profile: ExtendedProfile) extends AccountControllerBase
  with AccountService with RepositoryService with ActivityService with OneselfAuthenticator with Profile
```
```
class ScalatraBootstrap extends LifeCycle {
  override def init(context: ServletContext) {
    val driver = Database.driver(context)
    context.mount(new IndexController(driver), "/")
    context.mount(new SearchController(driver), "/")
    ...
```

### DIFF
--- a/src/main/scala/ScalatraBootstrap.scala
+++ b/src/main/scala/ScalatraBootstrap.scala
@@ -1,4 +1,4 @@
-import _root_.servlet.{BasicAuthenticationFilter, TransactionFilter}
+import _root_.servlet.{BasicAuthenticationFilter, Database, TransactionFilter}
 import app._
 import jp.sf.amateras.scalatra.forms.ValidationJavaScriptProvider
 import org.scalatra._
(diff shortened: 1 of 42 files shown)


takezoe:
Merged to #791

PR #306, opened 2014-03-06
Display user's full name in header link

olivierdagenais:
We're recording users' full names but aren't showing it to them in the header.  This pull request remedies this.
## Manual testing
1. Sign in as **root**
2. Edit the profile and set the "Full Name" field to **I am root**
3. Sign out.
4. Sign in as **root** again.
5. Look in the header: i am root

Mission accomplished!


uli-heller:
I'd like to see this pull request getting merged. Is this an option, now that we try to do almost everything different than GitHub?

At my site, many users have login ids like "x112245" and it is pretty hard to remember all these names :(

PR #356, opened 2014-04-29
Display SSH URLs at first

okumin:
The availability to access repositories by SSH is a fine feature.
But in some pages, gitbucket displays HTTP's URL only or at first.
Maybe, users who enable SSH access prefer web pages to display SSH URLs.


purpleraven:
+1

PR #540, opened 2014-11-04
add current user to notifications

rlazoti:
Add the message's author and improve a little bit the notification content to be more informative. :)


takezoe:
In GitHub, the author is displayed as From alias of notification email. We should follow it.

takezoe:
Since the notification has been separated to [gitbucket-notifications-plugin](https://github.com/gitbucket/gitbucket-notifications-plugin), I close this issue. If you have any request about the notification, feel free to ask at the notification plugin project.

Or, the notification has been pluggable, so you can create your own notification plugin instead of the default one.

PR #586, opened 2015-01-02
add icon on each diff header

nazoking:
Add delete/add icon on each diff header.
Because it is difficult to be discriminated binary file has been added or deleted.
5d34e6c0dccc58f91bec83d61b33fb1c

↓

b2c0ff385bbd30bf9c4dccda869732cb


takezoe:
@nazoking Good pull request, thanks. We want to add diff indicator like GitHub actually :-)

PR #594, opened 2015-01-09
Easily extended systems as greasemonkey / user scripts.

nazoking:
Administrator set site-script url, and footer of all pages has this url as script src.
Site script will work in the same way as Greasmonkey.

Administrator if you have their own repository, it can be used as a script file storage that you can easily manage the history and permissions.


takezoe:
@nazoking It's a good idea to extend GitBucket easily! But test compilation is failed. Could you fix it?

nazoking:
sorry, i fix it.

takezoe:
@nazoking We can substitute the new plug-in system for this feature. New plug-in system comes in the next release 2.9. Check this example:
https://github.com/takezoe/gitbucket-gist-plugin/blob/1a698390ba383b23c64c6cc4c8608e2f3863a7bf/src/main/scala/Plugin.scala#L18-L21

nazoking:
out of date

PR #597, opened 2015-01-12
UI usability fix on Issue comment and wiki editor

pibi:
home - piko-placeholder 2015-01-12 12-42-16
Avoid confusion on the issue page.
test - issue 1 - piko-placeholder 2015-01-12 12-11-31


pibi:
What do you think about a Cancel button between Edit and New Issue on top of the Issue Editor?

takezoe:
Thanks for request.
However GitBucket is according to GitHub UI basically. So we can't take PR which modifies it like this.

PR #604, opened 2015-01-20
Enable matching URLs end with slash

johnthemountainy:
Hi.
If we make a request to http://localhost:8080/username/, we will receive 404 status code.
This PR supports end slash in some urls.

Thank you for good product!

### DIFF
--- a/src/main/scala/app/AccountController.scala
+++ b/src/main/scala/app/AccountController.scala
@@ -97,7 +97,7 @@ trait AccountControllerBase extends AccountManagementControllerBase {
   /**
    * Displays user information.
    */
-  get("/:userName") {
+  get("/:userName/?") {
     val userName = params("userName")
     getAccountByUserName(userName).map { account =>
       params.getOrElse("tab", "repositories") match {
--- a/src/main/scala/app/IssuesController.scala
+++ b/src/main/scala/app/IssuesController.scala
@@ -48,7 +48,7 @@ trait IssuesControllerBase extends ControllerBase {
       "content" -> trim(optional(text()))
     )(IssueStateForm.apply)
 
-  get("/:owner/:repository/issues")(referrersOnly { repository =>
+  get("/:owner/:repository/issues/?")(referrersOnly { repository =>
     val q = request.getParameter("q")
     if(Option(q).exists(_.contains("is:pr"))){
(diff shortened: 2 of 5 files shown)


geek1011:
Any news on this?

nigelsim:
Any chance this can get merged soon? This is an issue for us as the Jenkins GitHub Plugin appends a slash to the URL supplied to it. Additionally, because the 404 page for GitBucket is the builtin Jetty one it is confusing to users when they hit these URLs.

tundratim:
Any further progress on this issue?

vidkun:
Any updates on when we can get a fix for this trailing slash issue? This is also breaking our Jenkins integration as well.

tundratim:
Any idea on when this will go into a release?

lefou:
It can be frustrating sometimes, when a maintainer doesn't comment.

tundratim:
@lefou  Well ... the maintainers are pretty much all volunteering their time, so I don't think being cross with them is warranted.  I just would like an ETA for this fix even if it's 10 months out or whatever.

takezoe:
Sorry, I couldn't mention this for a long time because I thought that supporting URL with `/` partially is not good idea. If we support it, we should support in GitBucket as a whole. And we should be carefully to accept this change because if URL specification is changed once, we can't change it easily in the future.

Anyway, supporting URL with `/` to help Jenkins integration is meaningful. I would like to advance this topic.

McFoggy:
@takezoe not clear to me in #1840 who is adding the trailing '/'. Is it jenkins or nginx?

> And we should be carefully to accept this change because if URL specification is changed once, we can't change it easily in the future.

This is perfectly understandable but IMO http browsing on:
- `GITBUCKET/username` is probably exactly the same as `GITBUCKET/username/`
- `GITBUCKET/username/repository` is also probably the exact same as `GITBUCKET/username/repository/`

vidkun:
@McFoggy From what I can tell, it appears to be jenkins that is adding the '/'. If it were nginx, I'd expect it to be added when trying to directly browse to a URL as well. However, we are only seeing it on the jenkins integration.

takezoe:
I created an another pull request #1921 that makes possible to support path which ends with slash for all actions.

takezoe:
Fixed in #1921

PR #605, opened 2015-01-20
Add Connection Pool

rlazoti:
As some people reported [here](https://github.com/takezoe/gitbucket/issues/463) and [here](https://github.com/takezoe/gitbucket/issues/460) and I realized it as well, GitBucket has some performance problems.

This pull request add a connection pool using the [c3p0](https://github.com/swaldman/c3p0) library and [typesafe config](https://github.com/typesafehub/config) as configuration library.
I'm using (about 10 concurrent users and 60 repositories) this branch internally for more than 2 months with no performance issue so far.

The connection pool (c3p0) is using its [default settings](http://www.mchange.com/projects/c3p0/) but it can be [changed directly in the database.conf file](http://www.mchange.com/projects/c3p0/#c3p0_conf) if needed.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -49,6 +49,8 @@ object MyBuild extends Build {
       "org.eclipse.jetty" % "jetty-webapp" % "8.1.8.v20121106" % "container;provided",
       "org.eclipse.jetty.orbit" % "javax.servlet" % "3.0.0.v201112011016" % "container;provided;test" artifacts Artifact("javax.servlet", "jar", "jar"),
       "junit" % "junit" % "4.11" % "test",
+      "com.mchange" % "c3p0" % "0.9.5",
+      "com.typesafe" % "config" % "1.2.1",
       "com.typesafe.play" %% "twirl-compiler" % "1.0.2"
     ),
     EclipseKeys.withSource := true,
--- a/src/main/resources/database.conf
+++ b/src/main/resources/database.conf
@@ -1,0 +1,6 @@
+db {
+  driver = "org.h2.Driver"
+  url = "jdbc:h2:${DatabaseHome};MVCC=true"
+  user = "sa"
+  password = "sa"
+}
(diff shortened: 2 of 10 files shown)


takezoe:
@shimamoto Could you review this pull request?

hansgru:
This is very nice, but the real performance comes from the Tomcat connection pool:
http://tomcat.apache.org/tomcat-7.0-doc/jdbc-pool.html

The only thing that GitBucket needs to do is to use a JNDI connection (or offer it as an option)

The Tomcat connection pool can be used with Jetty too.

rlazoti:
@hansgru what do you mean by 'real performance'?

You said the container-manager pool (JNDI) has real performance, so could you explain or add any articles that show why an application pool (c3p0) has no real performance?

hansgru:
> You said the container-manager pool (JNDI) has real performance

Nope :) . I said that the **Tomcat** Connection Pool implementation (that can be used also outside of Tomcat, but when you deploy in Tomcat it's there) is really fast :) .

One can use a "container managed pool) in any web application with the help of JNDI (so if GitBucket would use JNDI, than users could configure in Tomcat so that the JNDI connection **points** to that fast Tomcat Connection Pool). In the case of Jetty or Servlet Containers, a little more work needs to be done, but you can deploy the Tomcat Connection Pool in any other Servlet Container too.

> so could you explain or add any articles that show why an application pool (c3p0) has no real performance?

http://www.tomcatexpert.com/blog/2010/03/22/understanding-jdbc-pool-performance-improvements

c3po was always just a toy, and even if it's deployed default with Hibernate, it was always mentioned not to use it for production purposes.

shimamoto:
@rlazoti Thank you for this PR. We merge this PR because it is useful modification for now gitbucket.

In the future, We are planning to upgrade to Slick 3. So we might change HikariCP including built-in support for Slick 3.

MarcusMoeller:
@rlazoti could you please come up with an example config? I am currently using apache as ajp proxy like:

  ProxyPass /gitbucket ajp://localhost:8009/gitbucket

Tomcat itselfs only listens on localhost 8009. How can I change the setup to a ConnectionPool?

PR #615, opened 2015-01-31
fix/xss by raw html

nazoking:
test/html is cause of xss

### DIFF
--- a/src/main/scala/app/AccountController.scala
+++ b/src/main/scala/app/AccountController.scala
@@ -135,8 +135,9 @@ trait AccountControllerBase extends AccountManagementControllerBase {
   get("/:userName/_avatar"){
     val userName = params("userName")
     getAccountByUserName(userName).flatMap(_.image).map { image =>
-      contentType = FileUtil.getMimeType(image)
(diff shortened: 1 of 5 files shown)


takezoe:
@nazoking Could this problem be fixed by adding a conversion from `text/html` to `text/plain` into `FileUtil#getMimeType()` as below?

```scala
def getMimeType(name: String): String =
  defining(URLConnection.getFileNameMap()){ fileNameMap =>
    fileNameMap.getContentTypeFor(name) match {
      case null        => "application/octet-stream"
      case "text/html" => "text/plain" // Add this line
      case mimeType    => mimeType
    }
  }
```

nazoking:
I thought that `FileUtil#getMimeType()` is sometimes used in other than the output of raw files, Impact is too large. But you know better than me about this project code, so I will follow your instructions.

But we must add a special http header `X-Content-Type-Options: nosniff`.for IE (even this is not perfect..).

XSS caused by content-type is very difficult. The Github dose not use same domain for raw output, but same way is raise the bar of install...

takezoe:
@nazoking I understood. Thanks.

Issue #424, opened 2014-07-04
Unable to fork projects


katetem:
After selected the fork action, the project won't get forked, instead nothing happens, and was forcing me to return back to the project index.
GitBucket Version: 2.0.0

takezoe:
Maybe this issue has been fixed already.

Issue #472, opened 2014-08-14
Some buttons are not responding


ziggystar:
In Firefox (where I do have some privacy plugins like Ghostery, Noscript, AdBlock installed but they should be disabled on my site), the following buttons do not respond:
- "Click to create a pull request for this comparison" has never worked (it's working in Chromium)
- "Edit" to select branches for new pull request has only worked once or twice (works in Chromium)
- the login button on the front page (next to the news list) doesn't work in Firefox, nor in Chromium

Does anyone know why the buttons on the pull-request page are that problematic?

With my limited JS knowledge, I was able to locate the following code

```
  $('#show-form').click(function(){
    $(this).hide();
    $('#pull-request-form').show();
  });
```

When I execute in the JS console `$('#show-form')`, I retrieve the button for creating the pull request.
When I execute `$('#pull-request-form').show()`, then the form is shown (which doesn't happen when clicking it). It appears, that somehow the code to install this handler is never executed.

I noticed that the JS code inside the HTML file is after the closing `</body></html>` tag. Is this legal?

ziggystar:
I'm getting the same JS error with other buttons, and also when using drop-down lists to assign users to issues. There are really many interface components that don't work in Firefox.

takezoe:
Close because this issue looks out of date.

PR #662, opened 2015-03-21
Fork should work regardless of group membership

calid:
Fork button had no effect if you didn't belong to a group.

This may explain #424

### DIFF
--- a/src/main/twirl/gitbucket/core/menu.scala.html
+++ b/src/main/twirl/gitbucket/core/menu.scala.html
@@ -43,11 +43,7 @@
       @if(loginAccount.isEmpty){
         <a title="You must be signed in to fork a repository" href="@path/signin" class="btn btn-small" style="margin-bottom: 10px;">Fork</a>
       } else {
-        @if(isNoGroup) {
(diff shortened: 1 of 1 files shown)


takezoe:
What browser (and version) are you using? I can't reproduce this issue the latest version of Chrome on Mac. And expected behavior for non-group uses is fork the repository immediately, not open facebox.

takezoe:
I expect that following JavaScript does not work on some browsers:
https://github.com/takezoe/gitbucket/blob/d495b04d859b637830349c17f028c58df3b7f59f/src/main/twirl/gitbucket/core/menu.scala.html#L197-L210

calid:
@takezoe Firefox on both windows and linux.  Range of versions.

calid:
I'll revert the facebox change, and play around with the JS.  If there is alternative JS you want me to try let me know.

calid:
Also looks like #472 is related

calid:
@takezoe fixing this is going to be a little over my head.  That being the case I've opened a standalone issue #677 and I'm going to go ahead and close this PR.

PR #693, opened 2015-04-02
Improve searching with Lucene

nus:
Improve speed of the searching with Lucene.

Note that it is slow to search when the first time searching or after the repository was modified because the processing contains indexing.

The index data will store in `$GITBUCKET_HOME/tmp/_index` directory.


dkowis:
:+1: Searching is pretty terrible with a repo with lots of files. 

Could it perhaps update the index on a push?

nus:
How can I fix this failed?
Only push with any changes?

takezoe:
We are planning embedded Elasticsearch integration for full-text search.

nus:
@takezoe
I am looking forward to the full-text search!

PR #697, opened 2015-04-06
let non-anonymous user read access to public repos with basic auth

banjun:
In GitBucket 3.1, http (read) access to public repos with basic authentication fails, when the **Anonymous access** setting is set to **Deny**.
They should have read/write access permission to public repos with the correct basic auth headers.


nus:
:+1:  Thanks

We have the same problem.

Issue #689, opened 2015-03-31
Mailer#toNotify's Future is likely to be executed before the transaction is committed.


troter:
When create issue or open new pull request, gitbucket will try to send a notification.
But currently, It may not be send notification. Because Servlet Filter's database session and Mailer's database session are different, and Mailer#toNotify's Future is likely to be executed before the transaction is committed.

This is a dirty patch.
```diff
diff --git a/src/main/scala/gitbucket/core/util/Notifier.scala b/src/main/scala/gitbucket/core/util/Notifier.scala
index 9e80917..27fcf9f 100644
--- a/src/main/scala/gitbucket/core/util/Notifier.scala
+++ b/src/main/scala/gitbucket/core/util/Notifier.scala
@@ -73,6 +73,7 @@ class Mailer(private val smtp: Smtp) extends Notifier {

     val f = Future {
       database withSession { implicit session =>
+        Thread.sleep(10*1000)
         getIssue(r.owner, r.name, issueId.toString) foreach { issue =>
           defining(
               s"[${r.name}] ${issue.title} (#${issueId})" ->
```

takezoe:
@shimamoto Do you have any idea to solve this issue?

rlazoti:
@takezoe I've already fixed it, I'm going to send you a PR. :)

takezoe:
Fixed in #699

PR #699, opened 2015-04-07
Fix issue notifications

rlazoti:
As explained by @troter on Issue #689, the Notification class can't notify (send email) when a new issue or pull request is created because the transaction wasn't committed yet, and the getIssue method inside the Notifier class runs in a different session.

To fix it I just pass the issue object instead of issueId to the Notifier class.


takezoe:
@shimamoto Please review this pull request!

PR #702, opened 2015-04-08
Fix basic auth read when anon access disabled

marklacroix:
Basic authentication fails when not updating a public repository with the "Disable Anonymous Access" option enabled


marklacroix:
Oops, missed #697 before working on this. Leaving it open as an alternate solution, but feel free to close in favor of #697

PR #714, opened 2015-04-15
Patch version identifier and Account display mailAddress 

mann-ed:
This pull request is adding in patchVersion so that the releases like 3.1.1 will display the full release in the header. It will work with 3.2 or 3.2.1 versions style.

This also has a change to display the mailAddress under account.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ project/plugins/project/
 # IntelliJ specific
 .idea/
 .idea_modules/
+/bin/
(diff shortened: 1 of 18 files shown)


takezoe:
Please divide this pull request into Patch version identifier and Account display mailAddress.

About displaying mail address, I assume users should be possible to select whether or not to display their own e-mail address like GitHub. At least, mail address should be displayed for only logged-in users.

PR #727, opened 2015-04-24
Adding the possibility to have read only collaborators

mericano1:
Hi, I have changed the code a bit to allow read only collaborators.
That means you can only fork the repo but not push to it. Initially I was thinking to create a role based auth, it would be much more flexible, but this is what I have so far. 

Let me know what you think


zabalit:
Hi, is this thread stuck? Read-only collaborators would be a nice feature.

mericano1:
It was a long time ago, would probably require some work now. The saddest thing is that I never had any feedback on the changes..

monkiki:
Any news about this feature?

aadrian:
@mericano1 @zabalit @monkiki GitBucket has "Guest Collaborators" now.
Does that feature fulfill the same role as this "read only" ? Can you achieve the same thing?

Issue #441, opened 2014-07-23
gitbucket is slow


Centaur:
We throw gitbucket.war into tomcat on one of our servers. This app is obviously slower than others in the same tomcat instance, every click seems cost 1+ second to respond.

nazoking:
I think, acitivity table is too large.
My gitbucket server has 377 reposotory, 21577 activity from 2013-10-28 08:52:43.768.
ActivityService.getRecentActivities query (that is used in the top page), take 15945 ms until the end .

takezoe:
It might be to delete the old activity records automatically.

takezoe:
We will add a new configuration item to specify limit of activity logs in GitBucket 3.2. If your activity table is too large, please configure it at the system settings console.

takezoe:
Fixed in #732

MarcusMoeller:
Could you please tell me which value this is? Is it the number of recent activity items displayed on the dashboard? What is a good value ? e.g. 10, 100 ?

PR #732, opened 2015-04-29
(refs #441) Add limit of activity log configuration

takezoe:
Delete old activity logs if activity log limit has been specified.


takezoe:
@shimamoto Please review this pull request because this pull request contains new library dependencies.

src/main/scala/gitbucket/core/service/ActivityService.scala
```diff
@@ -7,6 +7,12 @@ import profile.simple._
 
 trait ActivityService {
 
+  def deleteOldActivities(limit: Int)(implicit s: Session): Int = {
+    Activities.map(_.activityId).sortBy(_ desc).drop(limit).firstOption.map { id =>
+      Activities.filter(_.activityId < id.bind).delete
```
shimamoto: In this case, the limit + 1 size data are remained.

shimamoto:
LGTM

PR #794, opened 2015-06-27
doc change: build.xml -> release/build.xml

uli-heller:
The doc references a top level build.xml, which doesn't exist any more.


uli-heller:
This is a mess, please ignore...

PR #796, opened 2015-06-27
Collect all recipients into the TO field and send out the notification emails only once to all of them

uli-heller:
Probably not useful to anyone but me ;)

See doc/README.combined-mail.md for details.

### DIFF
--- a/doc/README.combined-mail.md
+++ b/doc/README.combined-mail.md
@@ -1,0 +1,46 @@
+Combined Mail
+=============
+
+As of version 3.4, gitbucket sends out individual mails to each recipient.
+This way, one recipient doesn't know about the other recipients.
+
+For me, I'd like to see a list of all recipients. This way, I can verify
+whether a certain person got a notification of not.
+
+So I've created a patch which collects all recipients within the "to"
+list and sends the mail once to all of them.
+
+To verify this, you can:
+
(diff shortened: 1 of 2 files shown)


shimamoto:
Thanks PR.

The current implementation is in line with GitHub's behavior.
So we can **not** take in this fix as the **default** behavior.

I think that it is good if as an optional feature.
Though, if there is tens of people, include all into TO?
Do you have any advice?

uli-heller:
Hi,

I agree it probably doesn't make sense to do this. I personally need the feature to find out which mail gets send to whom. So it is kind of a debugging aid for me personally.

A mail header field like "TO" doesn't have any size limitation as far as I know. So you can technically add hundreds of recipients into the TO field. Some mail readers may not like this, though.

If you don't want the people to see each others mail addresses, some people use the BCC field instead of the TO field.

I don't know what GITHUB does, but to mee this would probably sound reasonable:

* TO ... the assignee
* CC/BCC ... all the others

Best regards, Uli

takezoe:
@uli-heller I think GitBucket users have each own request about email notification, so we want to consider some options about email notification behaviour carefully. Maybe customizable by plug-in is one of these options.

uli-heller:
@takezoe: Thanks for looking at the issue. I'd really like to see a plugin mechanism that's capable in making something like this customizable. I might even contribute ;)

Issue #738, opened 2015-05-02
Using [raw] button: very slow or timeout


3-O:
Upon attempting to display a 124kB svg file raw, I invariabily get a "Bad gateway" error (behind proxy) and the file is never displayed raw.
Similarly, displaying a text file raw is incredibly slow.
Is this the reason why github is serving such files from a different server?
https://raw.githubusercontent.com/takezoe/gitbucket/3732963d4baecc0f0ee790fa11d64c939083ea58/etc/icons.svg

Bug?

3-O:
java.lang.IllegalStateException: STREAM
        at org.eclipse.jetty.server.Response.getWriter(Response.java:693)
        at org.scalatra.ScalatraBase$class.renderUncaughtException(ScalatraBase.scala:214)
        at gitbucket.core.controller.ControllerBase.renderUncaughtException(ControllerBase.scala:27)
        at org.scalatra.ScalatraBase$$anonfun$executeRoutes$2$$anonfun$apply$5.apply(ScalatraBase.scala:182)
        at org.scalatra.ScalatraBase$$anonfun$executeRoutes$2$$anonfun$apply$5.apply(ScalatraBase.scala:179)
        at org.scalatra.ScalatraBase$class.org$scalatra$ScalatraBase$$cradleHalt(ScalatraBase.scala:206)
        at org.scalatra.ScalatraBase$$anonfun$executeRoutes$2.apply(ScalatraBase.scala:176)
        at org.scalatra.ScalatraBase$$anonfun$executeRoutes$2.apply(ScalatraBase.scala:175)
        at org.scalatra.ScalatraBase$class.org$scalatra$ScalatraBase$$cradleHalt(ScalatraBase.scala:206)
        at org.scalatra.ScalatraBase$class.executeRoutes(ScalatraBase.scala:175)
        at gitbucket.core.controller.ControllerBase.executeRoutes(ControllerBase.scala:27)
        at org.scalatra.ScalatraBase$$anonfun$handle$1.apply$mcV$sp(ScalatraBase.scala:113)
        at org.scalatra.ScalatraBase$$anonfun$handle$1.apply(ScalatraBase.scala:113)
        at org.scalatra.ScalatraBase$$anonfun$handle$1.apply(ScalatraBase.scala:113)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
        at org.scalatra.DynamicScope$class.withResponse(DynamicScope.scala:80)
[...]

PR #828, opened 2015-07-12
fix: wrong content type

sekia556:
In the repository viewer, Gitbucket returns Content-type as "application/octet-stream", even if the content is an image (ex. jpeg, png).
So, some browsers (at least IE) cannot display the images. (Chrome and Firefox can display the image even if Content-type is "application/octet-stream").

It seems that all the raw data are returned as "application/octet-stream" in RepositoryViewerController.scala.
I think that every content should be returned with the correct MIME-type.


takezoe:
Current behavior was introduced to fix #738. This issue is caused by Scalatra bug. It will be fixed in the next version of Scalatra, so we can revert #738 after upgrade Scalatra.

sekia556:
Thank you for quick response.
OK, I'll wait to fix #738.

takezoe:
Waiting for Scalatra 2.4.

takezoe:
#738 was reverted in GitBucket 3.10

Issue #820, opened 2015-07-09
Truncate long commit messages


trymnilsen:
Hi there

A small suggestion for improvement would be adding some small snippets of CSS to truncate long commit messages. 

Example
![image](https://cloud.githubusercontent.com/assets/1550685/8591178/caf46458-2626-11e5-9b8d-b32cab02246d.png)

How github does it (not that we need to directly copy it).
![image](https://cloud.githubusercontent.com/assets/1550685/8591199/ec8e928c-2626-11e5-98cd-212c9ba0e730.png)

McFoggy:
@takezoe this issue has been solved via PR and c

PR #846, opened 2015-07-29
solve #820 too long messages

McFoggy:
small enhancement to shorten long commit messages in the file list of repositories (#820)

image


garygreen:
Why do this in scala? It would be better to do it in CSS, the same way github does. This would make it easier if Gitbucket changes to be more responsive in the future as well.

takezoe:
@garygreen Agree. Our current implementation does it in Scala, but I think it's better to do it in CSS. I will merge this PR currently. Could you fix it in the future?

garygreen:
Don't merge. I'll fix it in CSS and do another pr.

takezoe:
@garygreen Good. I'm looking forward to your PR.

garygreen:
PR #882

takezoe:
Fixed in #882

Issue #838, opened 2015-07-21
Link to commits in other repositories


McFoggy:
Current reference linking fails when using the github syntax described [there](https://help.github.com/articles/writing-on-github/#references).

In gitbucket, in project GROUP1/Project1, referencing for example in an issue `Group2/Project2@SHA1` creates a link to `GROUP1/Project1/commit/SHA1` which is wrong and leads to the following error when clicked

```
org.eclipse.jgit.errors.MissingObjectException: Missing unknown 39978bada1e0b5625df5a1585eb2c8dd6899e141
    at org.eclipse.jgit.internal.storage.file.WindowCursor.open(WindowCursor.java:148)
    at org.eclipse.jgit.lib.ObjectReader.open(ObjectReader.java:229)
    at org.eclipse.jgit.revwalk.RevWalk.parseAny(RevWalk.java:839)
    at gitbucket.core.util.JGitUtil$.getRevCommitFromId(JGitUtil.scala:154)
    at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$25$$anonfun$apply$53$$anonfun$apply$54.apply(RepositoryViewerController.scala:347)
    at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$25$$anonfun$apply$53$$anonfun$apply$54.apply(RepositoryViewerController.scala:346)
    at gitbucket.core.util.ControlUtil$.using(ControlUtil.scala:26)
    at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$25$$anonfun$apply$53.apply(RepositoryViewerController.scala:346)
    at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$25$$anonfun$apply$53.apply(RepositoryViewerController.scala:343)
    at gitbucket.core.util.ReferrerAuthenticator$$anonfun$gitbucket$core$util$ReferrerAuthenticator$$authenticate$1$$anonfun$apply$12.apply(Authenticator.scala:123)
    at gitbucket.core.util.ReferrerAuthenticator$$anonfun$gitbucket$core$util$ReferrerAuthenticator$$authenticate$1$$anonfun$apply$12.apply(Authenticator.scala:121)
    at scala.Option.map(Option.scala:146)
    at gitbucket.core.util.ReferrerAuthenticator$$anonfun$gitbucket$core$util$ReferrerAuthenticator$$authenticate$1.apply(Authenticator.scala:121)
[...]

PR #847, opened 2015-07-29
Fix issue #838

officer:
Fix issue #838 
Simply response "NotFound" if specified repository was missing.
As long as I confirmed, This behaviour is the same as GitHub's.


McFoggy:
I am not sure you got the point. 
Even if answering a 404 is better than throwing an exception, the issue relates to the fact that the produced reference link is wrong ; not that the repositories/SHA1 do not exists.
Your commit from my understanding will provide better handling of wrong links but it will not prevent from creating those bad links.

officer:
Thank you for your quick response.
Your comment is definitely correct.

I will retry it..

PR #848, opened 2015-07-29
Increase container width

garygreen:
Matches Github's container width of 980px.


garygreen:
Thanks @takezoe. Would you be able to take a look at PR #849 ? All that does is basically make it easier to click the buttons.

takezoe:
OK. I will check it.

PR #857, opened 2015-08-03
small fixed: env.sh has been already removed.

ssogabe:


officer:
env.sh has not been removed but moved to release/ directory.

takezoe:
@officer Thanks for your comment. You are right. But sbt execution does not require env.sh no longer.

Issue #852, opened 2015-07-30
Add an announcement mechanism


McFoggy:
It would be great to have a way to notify users/groups for some announces.

Announces could:

- be persistent or not
- be notified by email or not
- have a scope target
    - global
    - list of group
    - list of individuals
- ...

For persistent announces, we could imagine to provide some panels in different places showing the announces relatively to the place we are on the site: main page, project page.

One of the first usage would be to announce: shutdown, upgrade, restart in advance to gitbucket users.

Such a mechanism could also serve as a basis for a kind of messaging system between individuals/groups.

_PS: of course we can split this issue in several smaller ones in order to deliver something step by step._

This could be implemented as a plugin also, but I am not sure we have ways today to add content from a plugin into existing gitbucket pages.

takezoe:
GitBucket has a simple notification panel at the top of right sidebar. Yes, I guess the more rich notification system you mentioned should be provided as plugin because notification requirement depends on usecase. We will provide more practical document about plugin development.

McFoggy:
> GitBucket has a simple notification panel at the top of right sidebar
Other than the flash notification? 
I updated on 3.4 and didn't saw any new panel for notification.

If the plugin mechanism allows enough extensibility (meaning enrichment of existing pages) then it will be a perfect candidate to develop this functionality as a plugin. 
Meanwhile for my own needs I'll probably create a small page to be able to email all users when a shutdown/upgrade/restart is planned.

McFoggy:
For those interested and waiting for something better, I have created in my branch [add-email-announces](https://github.com/McFoggy/gitbucket/tree/add-email-announces) a "mailing" facility so that you can inform all your active users about a next shutdown/upgrade

![image](https://cloud.githubusercontent.com/assets/1119660/9006590/7d0c5c7c-3789-11e5-8bdc-19320f98693d.png)

Enter a subject, a text description and an email announce will be sent to all active users of the system. Emails are sent via usage of BCC email field.

McFoggy:
Rebased on #4a35b65c2c

takezoe:
Cool.

> If the plugin mechanism allows enough extensibility (meaning enrichment of existing pages) then it will be a perfect candidate to develop this functionality as a plugin. 

I expect it's possible to implement as a plug-in. However we might have to add some new extension point to make this plug-in. For example, now we can add an item to existing menu by JavaScript as GreaseMonkey script. I think it's better that we can add an item via plug-in API.

[...]

PR #861, opened 2015-08-05
give an id to system admin menu container to allow plugin extension

McFoggy:
Having an id on the container of admin links allows a plugin to enhance it.

This will be used for example by [McFoggy/gitbucket-announce-plugin](https://github.com/McFoggy/gitbucket-announce-plugin)

### DIFF
--- a/src/main/twirl/gitbucket/core/admin/menu.scala.html
+++ b/src/main/twirl/gitbucket/core/admin/menu.scala.html
@@ -4,7 +4,7 @@
   <div class="row-fluid">
     <div class="span3">
       <div class="box">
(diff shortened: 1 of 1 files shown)


takezoe:
@McFoggy This pull request is GitBucket way in the current version. Hoever I think it might be better that provide API to add items to some main extensible points such as global header or administration menu.

McFoggy:
@takezoe as commented in #852 both possibilities for extension (real API versus html/css consistency) can do the trick. 

I will adapt to the choices made in the project.

Allowing page extension with plugin provided HTML implies that the site HTML/CSS has to remain stable. And this does not depend on the way the extension point is exposed.

PR #862, opened 2015-08-05
list installed plugins in the system administration menu

McFoggy:
this contribution adds a new entry in the system admin menu to see which plugins are installed. It provides basic information on those plugins.

image


McFoggy:
and this is the view with #861 integrated, so that [McFoggy/gitbucket-announce-plugin](https://github.com/McFoggy/gitbucket-announce-plugin) plugged into the menu

image

PR #896, opened 2015-09-02
split SMTP & notification, fixes #893

### LINKED ISSUE #893: separate SMTP settings from notifications ones
In order to be able to use the [announce plugin](McFoggy/gitbucket-announce-plugin) independently from the notifications mechanism, it would be great to separate the notifications settings from the SMTP ones.

Doing so, we could declare a SMTP for reuse but do not activate notifications. 
It would also allow in the future to:

- enhance notifications settings by providing a dedicated section for fine grain setup
- allow to have more communication channels than just SMTP

McFoggy:
this PR brings a separation between the setup of a SMTP server and the usage of notifications.

For the ones who have disabled notifications because they are too verbose in the current version _(or not enough flexible)_, it allows them to still define SMTP settings and use [announce plugin](https://github.com/McFoggy/gitbucket-announce-plugin) as expressed in McFoggy/gitbucket-announce-plugin#5

image

### DIFF
--- a/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala
+++ b/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala
@@ -24,7 +24,8 @@ trait SystemSettingsControllerBase extends ControllerBase {
     "activityLogLimit"         -> trim(label("Limit of activity logs", optional(number()))),
     "ssh"                      -> trim(label("SSH access", boolean())),
     "sshPort"                  -> trim(label("SSH port", optional(number()))),
-    "smtp"                     -> optionalIfNotChecked("notification", mapping(
+    "useSMTP"                 -> trim(label("SMTP", boolean())),
+    "smtp"                     -> optionalIfNotChecked("useSMTP", mapping(
         "host"                     -> trim(label("SMTP Host", text(required))),
         "port"                     -> trim(label("SMTP Port", optional(number()))),
         "user"                     -> trim(label("SMTP User", optional(text()))),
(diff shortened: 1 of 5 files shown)


McFoggy:
@takezoe can you please review the PR for inclusion to 3.7, I'd like to adapt the plugin for it's next release.

takezoe:
@McFoggy Concept of this PR seems good, I'd like to take it.

However it should be attention to usability little more. I think users can't know that SMTP configuration is required to send notification mail if these configuration will be separated. Therefore, we should add more kind validation and guide message. Can I do it?

McFoggy:
@takezoe I agree that we could document it better but as "mail" is the only notification channel available you'll notice that in the PR if you deactivate SMTP then  notification is also decativated automatically.
Of course in the future if other communication channels are added this behavior will have to be modified in favor of another mechanism ; a validation one why not.

takezoe:
@McFoggy I see. I will add only small notation (or UI modification) to this PR in order to guide users.

> Of course in the future if other communication channels are added this behavior will have to be modified in favor of another mechanism

Right. I think the notification way should be selectable if communication channel is pluggable.

McFoggy:
> I will add only small notation (or UI modification)
If you want I update the PR ; otherwise feel free of course to add such a notation.

> the notification way should be selectable if communication channel is pluggable.
I fully agree that it will be a required enhancement in the future.

takezoe:
@McFoggy Merged. I added small notation and migration in b11d36c3a5d66570cf7d8919d3c3faeab1ac2e2f.

McFoggy:
perfect thanks

PR #906, opened 2015-09-15
Truncate long repository names in the repository list

chazmuzz:
Truncate text inside a box-content-row div rather than flowing out of the box


takezoe:
@chazmuzz `div.box-content-row` is used in many places other than the repository list. In some places, truncate is not preferred. Therefore we can't merge this PR with no modification.

Is this PR for the repository list in the sidebar of the activity page? If true, we will add another new class to them.

chazmuzz:
Yes, it was for the sidebar in the activity page.  I saw that a couple of repository names are running out of the box.

takezoe:
@chazmuzz Thanks, I added small fix and merged.

PR #911, opened 2015-09-18
support sshPublicKey from LDAP

ccskoba:
LDAP server can provide user ssh publickey as sshPublicKey attribute.
it convenient, if PublicKeyAuthenticator support find publickey from LDAP.

I tried modify gitbucket and success it for me.

I did,
- add "SSH Public key attribute" field on System Setting LDAP section.  
  usually put "sshPublicKey"
- change PublicKeyAuthenticator
- add some method LDAPUtil

please advise these codes are effectual?

Thank you.　(Issue#910)

pubkeyfromldap

### DIFF
--- a/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala
+++ b/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala
@@ -44,6 +44,7 @@ trait SystemSettingsControllerBase extends ControllerBase {
         "additionalFilterCondition"-> trim(label("Additional filter condition", optional(text()))),
         "fullNameAttribute"        -> trim(label("Full name attribute", optional(text()))),
         "mailAttribute"            -> trim(label("Mail address attribute", optional(text()))),
+        "pubKeyAttribute"          -> trim(label("SSH Public key attribute", optional(text()))),
         "tls"                      -> trim(label("Enable TLS", optional(boolean()))),
         "ssl"                      -> trim(label("Enable SSL", optional(boolean()))),
         "keystore"                 -> trim(label("Keystore", optional(text())))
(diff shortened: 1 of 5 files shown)


takezoe:
@ccskoba We would like to merge #1104 which makes possible to connect via SSH as a generic user. I think it's incompatible with this request because it's unrealistic to look all LDAP entries to find a specific user's public key.
Unfortunately, I would like to close this pull request if you don't have any idea to achieve generic ssh user authentication and retrieving public key from LDAP.

ritschwumm:
how are public keys stored in ldap?
particularly, can one specific public key be encoded in more than exactly one way?

takezoe:
@ritschwumm It expects BASE64 encoded string as same as GitBucket DB. So we have to decode keys in LDAP using [SshUtil.str2PublicKey](https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/ssh/SshUtil.scala#L16) to compare with the requested key.

ritschwumm:
hm. for the generic user i could imagine to use a LIKE '% KEY %' query where KEY is pre-encoded base64, but can LDAP do that? (i didn't go for this because i wasn't sure whether i could rely on the format of the key always being 'type' 'octets' 'email')

ccskoba:
@takezo san, I have no idea to solve that compatibility issue.
please close this pull request.
thanks a lot.

ritschwumm:
so.. knowing a bit more about LDAP now i do think we can base64-encode the requested key, put it in a substring query like ```(sshKey=*389479834793475*) ``` (maybe including blanks left and right) and do a proper SshUtil.str2PublicKey  comparison afterwards. unless there are a lot of users with an sshKey attribute in LDAP this probably is performant enough.

eternalharvest:
I'm a just passing stranger who is considering to use gitbucket and require the LDAP authentication.
But I can't imagine the usecase of using publickey stored in LDAP server.

I think we can store same public key to the another 2+ user's public key attribute of LDAP entry, though I'm not sure if it true.
Public key itself can not be used as ID because it is not unique attribute in the LDAP database.
It is similar to the password itself cannot be used as a ID.

So, We **MUST NOT** manage public key at the place where we cannot ensure that the unique constraints are always satisfied.

I apologies if my thought is wrong, sorry.

Issue #787, opened 2015-06-25
Markdown Syntax


MarcusMoeller:
When I add a code block to README.md

```
$ make vagrant                   # build vagrant box
$ make vagrant-collectstatic-dev # collect all staticfiles
$ make vagrant-runserver         # run development server
```

it is formatted in one line instead of three.

hansgru:
have you used:
<pre>
```bash
$ make vagrant                   # build vagrant box
$ make vagrant-collectstatic-dev # collect all staticfiles
$ make vagrant-runserver         # run development server
```
</pre>
for your syntax?

MarcusMoeller:
Yes

MarcusMoeller:
Here is an example:
https://freie.software/adfinis-sygroup/freedomvote/blob/master/README.md?raw=true

hansgru:
you are missing blank lines before and after the code blocks.

See here: http://gitbucket.herokuapp.com/root/markdown/blob/master/README.md   (but it might disappear since it's a demo)

It is also recommended to specify what kind of syntax is used (python, bash, js, html) in the code block.

takezoe:
We will switch markdown processor to [markedj](https://github.com/gitbucket/markedj) from pegdown in GitBucket 3.7. Almost markdown syntax issues will be solved by markedj.

takezoe:
Merged to #912

Issue #210, opened 2013-11-25
Broken markdown


jtyr:
If I create a README.md with the following content:

---[ README.md START]---
```
cd /tmp
mkdir aaa
cd aaa
```
---[ README.md END]---

it shows the code block as one line.

takezoe:
We will switch markdown processor to [markedj](https://github.com/gitbucket/markedj) from pegdown in GitBucket 3.7. Almost markdown syntax issues will be solved by markedj.

takezoe:
Merged with #912

PR #912, opened 2015-09-21
Switch markdown processor to markedj from pegdown

takezoe:


takezoe:
This PR will solve also #210 and #787.

PR #920, opened 2015-10-01
fix pull-request url on webhook payload

nazoking:


takezoe:
@nazoking Thanks for your pull request!
I think that introducing a new case class `IssueOrPullRequest` is too much for this purpose.
Adding `Boolean` property to `ApiComment` and `ApiIssue` might be enough.
How do you think?

nazoking:
@takezoe Thanks review. I remove IssueOrPullRequest ( I thinked that api and html has difference of url for example `pull` -> `pulls`, `issues` -> `issues`, but it is my mistake :sweat: . )

PR #937, opened 2015-10-12
add getAccountByGroupName method

pinghe:
add getAccountByGroupName method

### DIFF
--- a/src/main/scala/gitbucket/core/service/AccountService.scala
+++ b/src/main/scala/gitbucket/core/service/AccountService.scala
@@ -90,6 +90,18 @@ trait AccountService {
   def getAccountByMailAddress(mailAddress: String, includeRemoved: Boolean = false)(implicit s: Session): Option[Account] =
     Accounts filter(t => (t.mailAddress.toLowerCase === mailAddress.toLowerCase.bind) && (t.removed === false.bind, !includeRemoved)) firstOption
 
+  def getAccountByGroupName(groupName: String, includeRemoved: Boolean = false)(implicit s: Session): List[Account] = {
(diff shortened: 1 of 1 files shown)


takezoe:
@pinghe What's the purpose of this added `getAccountByGroupName `?

pinghe:
for gitbucket-announce-plugin, i want email to group.

takezoe:
@pinghe Thanks for your response.
But we don't want to add new methods to services now because we are working for upgrade Slick to 3.x and it will make large impact to services API. Could you make your own new service that has `getAccountByGroupName`?

McFoggy:
@takezoe if people are asked to write their own services in plugins, do you want to offer a compatibility at the database level? 
Not offering an API at java/scala level, means that it has to be maintained at DB level? 
Isn't it cleaner to keep the control on the API instead of encouraging direct DB usage?

takezoe:
@McFoggy Basically, I think we should provide API compatibility and plugins. It's better than keeping DB level compatibility, However we are now working for Slick upgrading for multiple DB support other than H2. It can't avoid API changes. After upgrading to Slick3, we will provide stable API for plugins and it also provides abstraction layer that's possible to treat DBs other than H2.

Also, I think it is reasonable way that plugins have their own services if core API is not enough. We don't restrain direct DB usage in plugins. But it does not mean we will keep DB level compatibility forever. Of course, we will continue to make effort to keep both of API and DB compatibility (but priority is API level compatibility) as much as possible, but now is not that time.

McFoggy:
I am fine with such a rational decision as it is bound to well thought direction. Thanks for your input.

PR #939, opened 2015-10-12
Support uploading PDF files

nus:
Support uploading PDF files from a issue comment form.
See also https://github.com/blog/2061-attach-files-to-comments
and #12 .

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -56,6 +56,7 @@ object MyBuild extends Build {
       "org.apache.commons" % "commons-email" % "1.3.3",
       "org.apache.httpcomponents" % "httpclient" % "4.3.6",
       "org.apache.sshd" % "apache-sshd" % "0.11.0",
+      "org.apache.tika" % "tika-core" % "1.10",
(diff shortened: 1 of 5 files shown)


takezoe:
@nus Don't you need other file type (Word, Excel and PowerPoint) support?

imeszaros:
I think uploading more kind if popular files would be very useful..

txt, rtf, office files, etc..

nus:
Thanks for your comments.

@takezoe I prefer PDF to other Document formats. That's the reason why I selected a PDF.

I will add below file formats.
- [x] PDF .pdf
- [ ] Word .docx, .doc
- [ ] Excel .xlsx
- [ ] Powerpoint .pptx
- [ ] Text .txt
- [ ] RTF .rtf

Does anyone want to add other file formats?

takezoe:
@nus Please notice, GitBucket is according to GitHub specification basically. So we might reject implementation of more than over GitHub.

nus:
The following extensions are allowed for upload.

- .gif
- .jpg
- .png
- .docx
- .pptx
- .xlsx
- .txt

Issue #933, opened 2015-10-09
Unable to download large file


bertramn:
I checked in a large file (900mb) into the git repo and was trying to download it as raw content.

http://git.vagrant.v8:8080/gitbucket/root/test-big/blob/master/Fedora-Cloud_Atomic-x86_64-22.iso

but fails with below stack trace:

```
org.eclipse.jgit.errors.LargeObjectException: b8e8d679fa3c4f9b45aefd84cae1261aa99d04a7 exceeds size limit
	at org.eclipse.jgit.internal.storage.file.LargePackedWholeObject.getCachedBytes(LargePackedWholeObject.java:98)
	at org.eclipse.jgit.lib.ObjectLoader.getBytes(ObjectLoader.java:100)
	at gitbucket.core.util.JGitUtil$$anonfun$getContentFromId$1.apply(JGitUtil.scala:719)
	at gitbucket.core.util.JGitUtil$$anonfun$getContentFromId$1.apply(JGitUtil.scala:714)
	at gitbucket.core.util.ControlUtil$.using(ControlUtil.scala:17)
	at gitbucket.core.util.JGitUtil$.getContentFromId(JGitUtil.scala:714)
	at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$22$$anonfun$apply$44$$anonfun$apply$45$$anonfun$apply$46.apply(RepositoryViewerController.scala:296)
	at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$22$$anonfun$apply$44$$anonfun$apply$45$$anonfun$apply$46.apply(RepositoryViewerController.scala:293)
	at scala.Option.map(Option.scala:146)
	at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$22$$anonfun$apply$44$$anonfun$apply$45.apply(RepositoryViewerController.scala:293)
	at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$22$$anonfun$apply$44$$anonfun$apply$45.apply(RepositoryViewerController.scala:291)
	at gitbucket.core.util.ControlUtil$.using(ControlUtil.scala:26)
	at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$22$$anonfun$apply$44.apply(RepositoryViewerController.scala:291)
	at gitbucket.core.controller.RepositoryViewerControllerBase$$anonfun$22$$anonfun$apply$44.apply(RepositoryViewerController.scala:288)
[...]

PR #940, opened 2015-10-12
(ref #933) fix/Unable to download large file

nazoking:
I think we should be avoided that an exception ( separate issue from that gitbucket managing a large file with git ).
The cause of this exception is that FileUtil.isLarge has not compatibility with ObjectLoader.isLarge.

and I Think that

 1. JGitUtil.getContentFromId(git: Git, id: ObjectId, fetchLargeFile: Boolean) to deprecated
 2. `getContentFromId` split to 2 methods. `getObjectLoaderFromId` for large binary, and `getContentTextFromId` for small text contents.

This patch added only `getObjectLoaderFromId`.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
+++ b/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
@@ -298,7 +298,7 @@ trait RepositoryViewerControllerBase extends ControllerBase {
             contentType = "application/octet-stream"
             response.setContentLength(loader.getSize.toInt)
             loader.copyTo(response.getOutputStream)
-            Unit
+            ()
           } getOrElse NotFound
         } else {
           html.blob(id, repository, path.split("/").toList,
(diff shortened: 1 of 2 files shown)


src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
```diff
@@ -293,8 +293,12 @@ trait RepositoryViewerControllerBase extends ControllerBase {
...
         if(raw){
           // Download
-          JGitUtil.getContentFromId(git, objectId, true).map { bytes =>
-            RawData("application/octet-stream", bytes)
+          JGitUtil.getObjectLoaderFromId(git, objectId){ loader =>
+            //RawData("application/octet-stream", bytes)
+            contentType = "application/octet-stream"
+            response.setContentLength(loader.getSize.toInt)
+            loader.copyTo(response.getOutputStream)
+            Unit
```
takezoe: `Unit` references `object Unit`. Do you intend `()`?

takezoe:
@nazoking Looks good! But I added only one comment. Could you check it?

nazoking:
sorry. I fixed it.

PR #960, opened 2015-10-26
Creating repositories over API

jvamvas:
Implements two API endpoints imitating https://developer.github.com/v3/repos/#create:

`POST /user/repos` and `POST /groups/:group/repos`.

While Github's API uses `/orgs/:org/repos` for creating group repositories, the URL has been changed to `/groups/:group/repos` in order to adhere to Gitbucket's naming rules.

The following parameters are accepted:
* `name`. Required. The name of the repository
* `description`. A short description of the repository
* `private`. Whether the repository should be private. Default: `false`
* `auto_init`. Whether an initial commit with a README should be made. Default: `false`

If the pull request is merged, I will update the [Wiki entry](https://github.com/gitbucket/gitbucket/wiki/API-WebHook) accordingly.

### DIFF
--- a/src/main/scala/gitbucket/core/api/CreateARepository.scala
+++ b/src/main/scala/gitbucket/core/api/CreateARepository.scala
@@ -1,0 +1,19 @@
+package gitbucket.core.api
+
+/**
+ * https://developer.github.com/v3/repos/#create
+ * api form
+ */
+case class CreateARepository(
+    name: String,
+    description: Option[String],
+    `private`: Boolean = false,
+    auto_init: Boolean = false
+) {
+  def isValid: Boolean = {
+    name.length<=40 &&
+        name.matches("[a-zA-Z0-9\\-\\+_.]+") &&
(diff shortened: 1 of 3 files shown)


McFoggy:
@jvamvas I would suggest to provide consistency with github APIs. 
What about delivering both `/orgs/:org/repos` && `oups/:group/repos`? So that we are fully compatible with github API (and thus external tooling that could use this API) and also we are consistent with gitbucket naming.

takezoe:
@jvamvas Thanks for great pull request! 
`/orgs/:org/repos` might be better than  `/groups/:group/repos`. We will improve permission system to provide more detailed access controlling. So group will be similar to organization in GitHub in the future.

jvamvas:
@takezoe So I will change `/groups/:group/repos` to `/orgs/:org/repos` shortly.

takezoe:
@jvamvas Thanks! We will merge and release this request in 3.8.

PR #961, opened 2015-10-27
Update PullRequestsController.scala

superhj1987:
when the forked repository is the original repository(forkedRepository.repository.originRepositoryName&originUserName is None),then 404 will occur,so add the if to solve it.


takezoe:
@superhj1987 Does it mean the request source repository is not forked repository of the target repository but repository name is same by chance?

superhj1987:
@takezoe

1.UserA has a repository R1 -> UserA/R1.
2.UserB fork R1-> UserB/R1
3.UserA wants merge UserA/R1 to UserB/R1,the the fored repository is UserA/R1,but it has no originUserName and originRepositoryName,so Not found.

takezoe:
@superhj1987 Thanks. That's possible situation.

superhj1987:
Then, i add the if to solve the situation, but i don't know whether there is a better solution.

PR #992, opened 2015-11-14
fix lost web hook events on repository renamed.

nazoking:
sorry. I'd forgotten it.

Q. why you use delete-insert? not use update?


takezoe:
@nazoking Thanks for this fix!

> Q. why you use delete-insert? not use update?

To us? Because renaming repository requires to update some foreign keys.

PR #1010, opened 2015-12-16
support legacy route of List Statuses for a specific Ref api

nazoking:
because jenkins pull-request pulugin use it.


nazoking:
ref https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GHRepository.java#L859

PR #1039, opened 2016-01-09
Run H2 database as server mode

takezoe:


takezoe:
Close this request because GitBucket 4.0 will support MySQL and PostgreSQL as external database. If you want to run H2 as server mode, we recommend MySQL or PostgreSQL instead.

PR #1061, opened 2016-01-20
Fix styles

ShunsukeTadokoro:
- Remove unnecessary floating and re-layout sidemenu not to affect wiki textarea/preview etc
- Fix some FireFox v.40 layout collapse 

Firefox before

FireFox after


takezoe:
@ShunsukeTadokoro Thanks, Looks good! But mobile view is still not so good. 



However this is known issue before this request. So I merge this pull request. I'm grad if you continue to help us to solve issues about styles such as mobile view and following current GitHub UI.

Issue #965, opened 2015-10-30
Optimize the build artifacts.


aadrian:
Please optimize the build artifacts depending on how they will be used/deployed, in order to be more efficient and gain some performance too:

 - Compress JavaScript and CSS resources
 - Don't include the servlet api and jetty classes in the default WAR file what people will use to deploy to some other container like Tomcat, or GlassFish.
 - for running it directly a JAR file is enough (with embedded Jetty)
 - build a different WAR file with the DB in encrypted mode (for the users that need it).

ritschwumm:
with the second and third point, i guess you want to have two separate artifacts:
* a war file to deploy in a servlet container (without jetty classes)
* a single jar file including jetty and the webapp to run directly

i'd like this - i always found executable wars to be somewhat fishy. the easiest way to achieve this would imho be to split gitbucket into two separate subprojects: one for the pure war file, and one for the "runner" jar. for the latter, https://github.com/puniverse/capsule works quite well in my experience.

takezoe:
> Compress JavaScript and CSS resources

Exactly, we should do it.

> Don't include the servlet api and jetty classes in the default WAR file what people will use to deploy to some other container like Tomcat, or GlassFish.
> for running it directly a JAR file is enough (with embedded Jetty)

Many users who I know are running GitBucket by embedded Jetty. **Trying easy** is our first priority, so we are going to keep providing the executable war as the primary release artifact.

> build a different WAR file with the DB in encrypted mode (for the users that need it).

This is same as #68

PR #1065, opened 2016-01-22
issue #963: build executable war with sbt, fetch jetty jars at build time

### LINKED ISSUE #963: Jetty JARs should not be checked in but fetched at build.
The GitBucket repo seems to contain JAR files in the ```/embed-jetty``` directory.

These should be **fetched** at build time, not checked in, like the rest of the required JARs.

ritschwumm:
this should fix #963 and make the following files obsolete:
* /embed-jetty/*
* /release/build.xml

just use ```sbt executable``` to get an executable war.

### DIFF
--- a/executable.sbt
+++ b/executable.sbt
@@ -1,0 +1,77 @@
+val executableConfig	= config("executable").hide
+val executableKey		= TaskKey[File]("executable")
+
+Keys.ivyConfigurations		+= executableConfig
+libraryDependencies	++= Seq(
+	"org.eclipse.jetty"	%	"jetty-security"		% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-webapp"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-continuation"	% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-server"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-xml"				% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-http"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-servlet"			% "8.1.16.v20140903"	% "executable",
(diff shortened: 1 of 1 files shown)


executable.sbt
```diff
@@ -0,0 +1,77 @@
...
+libraryDependencies	++= Seq(
+	"org.eclipse.jetty"	%	"jetty-security"		% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-webapp"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-continuation"	% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-server"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-xml"				% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-http"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-servlet"			% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-io"				% "8.1.16.v20140903"	% "executable",
+	"org.eclipse.jetty"	%	"jetty-util"			% "8.1.16.v20140903"	% "executable"
```
takezoe: Extract Jetty version as one val

ritschwumm:
ok, added a constant. good enough for a merge?

 imho this is a bit of a hack anyway, see https://github.com/gitbucket/gitbucket/issues/965#issuecomment-174176388 so i wouldn't want to spend too much time on it. it is convenient, though.

takezoe:
@ritschwumm Looks good, great thanks. It seems to magic to me!! sbt is much difficult :-(

I'll merge this great request, but I don't know should we provide both of war file with and without Jetty classes. That makes new labor costs for us.

ritschwumm:
thanks for merging.
can you tell me what the assembly stuff in release/ is good for?
it would be pretty to do that in sbt, too.

takezoe:
@ritschwumm My hope about release operation is:
- Automate git tag and upload the war file and fingerprints to GitHub
- Also it's better to cover `deploy-assembly-jar.sh` with sbt

PR #1083, opened 2016-01-30
update sbt to newest version

ritschwumm:


ritschwumm:
stupid me - the sbt.jar is checked into git, so that would have to be updated, too.
considering how binaries can blow up a git repository, it might not be a good idea to update unless there's a real problem with the current version.

Issue #971, opened 2015-11-06
generic ssh user


semudu:
ssh urls are generated by the user name but I think, if we use a generic user for ssh (ex: git@gitbucket.machine.com), it would be better.

nus:
+1

nazoking:
+1

But this feature spec has some issues.

 - If some users already set same ssh-key to your gitbucket ?
   - raise exception on upgrade?
 - Can users login by current userid ?
   - If user `git` already exists ?

tette:
+1

In consideration of the integrity, I think this feature should be enabled in the options.

semudu:
i think, ssh-key is a special key for a user so same ssh-key is an extreme case. Of course, users can login by their userid but ssh clone url can be generated for one generic system user like "git". In this way, we don't have to create a system user on the host os for every user to login via ssh.

ritschwumm:
i want this. what's missing in @P1tt187 branch mentioned above?

ritschwumm:
just for the record: https://github.com/P1tt187/gitbucket/commit/4a02a880dfa3c31c7eb0edb509bfd01832981f62 does allow pulling, but not pushing. i think i know how to get that working, too.

ritschwumm:
i seem to have got git push working. now, what else is needed?
* instead of scanning all ssh keys manually, a findByKey should be used
* two accounts with different user names but the same ssh key cannot push at all at the moment
* an existing user named "git" should still be able to work
* how about legacy urls including the user name? i'd prefer getting rid of them in a clean sweep.

ritschwumm:
making this configurable in a proper way turns out to be slightly messy, see  #734

ritschwumm:
got it. as soon as #1084 is merged, i'll  open a pull request.

bviktor:
How does authentication work in this case? It's not quite clear to me and I can't find related documentation. Does username become "git" for EVERY user, then upon push/pull GitBucket searches for ALL user's SSH keys and if there's ANY match it authenticates?

ritschwumm:
almost. if the ssh username is "git" then ALL user's ssh keys are searched, and if there's EXACTLY ONE found, the user having that key is authenticated. if there's more than one matching key, authentication fails. if the username is anything else, only that user's key is used for authentication.

bviktor:
So the only reason I've found in this ticket for this weird behavior is being able to omit the username from the URL. Correct?

[...]

Issue #734, opened 2015-04-29
SSH base URL configuration should be separated from HTTP base URL


takezoe:

Cy4n1d3:
This would be great, I also need to run on a different port than 80 (or 8080) and thus I'm unable to enable SSH right now due to the port-stripping on base url.

ritschwumm:
not exactly trivial - the code using baseUrl and - derived from that - ssh urls is quite a mess.
i'm trying to get this cleaned up so i can proceed with #971. 

current plan:
* use a case class instead of a String for a baseUrl
* enhance this class to include an ssh-baseUrl
* move full-url generation and calculations in one place

does this make sense to you?

ritschwumm:
i guess this can be closed now

PR #1084, opened 2016-01-30
separate ssh url configuration from http baseurl

ritschwumm:
this should solve #734

the first few commits are mostly (hopefully) easily reviewable cleanups:
a lot of code could be made independent of the baseUrl by moving the generation of http/ssh-urls out of RepositoryInfo

after that, it was easy to make the hostname used for ssh urls configurable, additionally the views no longer need to know about the ssh port themselves.

note you still need to have a baseUrl configured in the settings for ssh because commit hooks need it to generate http urls.

i'd be glad if this got merged asap as it form the base for an implementation of #971 i have almost ready.

### DIFF
--- a/src/main/scala/gitbucket/core/service/RepositoryService.scala
+++ b/src/main/scala/gitbucket/core/service/RepositoryService.scala
@@ -2,7 +2,7 @@ package gitbucket.core.service
 
 import gitbucket.core.model.{Collaborator, Repository, Account}
 import gitbucket.core.model.Profile._
-import gitbucket.core.util.JGitUtil
+import gitbucket.core.util.{JGitUtil, RepoBase}
 import profile.simple._
 
 trait RepositoryService { self: AccountService =>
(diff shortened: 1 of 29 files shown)


ritschwumm:
@takezoe do i have a chance to get this merged?

the implementation of #971 would be quite bit messier without the
cleanups in here, but if you don't like it i can find other ways,
of course.

src/main/scala/gitbucket/core/service/RepositoryService.scala
```diff
@@ -389,31 +387,47 @@ trait RepositoryService { self: AccountService =>
...
-    def this(repo: JGitUtil.RepositoryInfo, model: Repository, forkedCount: Int, managers: Seq[String]) =
-      this(repo.owner, repo.name, repo.url, model, 0, 0, repo.commitCount, forkedCount, repo.branchList, repo.tags, managers)
+    def this(repo: JGitUtil.RepositoryInfo, model: Repository, 	forkedCount: Int, managers: Seq[String]) =
+      this(
+        repo.owner, repo.name, model,
+        0, 0,
+        repo.commitCount, forkedCount, repo.branchList, repo.tags, managers)
+  }
+
+  final class RepositoryUrls(repoBase:RepoBase, owner:String, name:String) {
```
takezoe: `RepositoryUrls` should be a case class rather than a `final` normal class.

src/main/scala/gitbucket/core/util/SshAddress.scala
```diff
@@ -0,0 +1,3 @@
+package gitbucket.core.util
+
+case class SshAddress(host:String, port:Int)
```
takezoe: Define this class in `SystemSettingsService` object.

src/main/scala/gitbucket/core/util/RepoBase.scala
```diff
@@ -0,0 +1,3 @@
+package gitbucket.core.util
+
+case class RepoBase(baseUrl:String, sshAddress:Option[SshAddress])
```
takezoe: `Context` has both of `baseUrl` and `settings.sshAddress`. Is this case class necessary? It's possible to create repository urls from only `Context`, and `Context` is also available at places where `repoBase` is used in this pull request.

takezoe:
@ritschwumm Yes, of course. This pull request seems to help some users. I added some comment. Could you check them?

ritschwumm:
thanks for looking into it, i'll try to address your comments asap

src/main/scala/gitbucket/core/service/RepositoryService.scala
ritschwumm: why? i think RepositoryUrl should do one and only one thing: provide URLs where you can access the repository. if you want the name, or the owner - go to the object whose job _that_ is, namely RepositoryInfo.

src/main/scala/gitbucket/core/util/SshAddress.scala
ritschwumm: will do

src/main/scala/gitbucket/core/util/RepoBase.scala
ritschwumm: yep, can be done - see 24b9a9a12c4e1192ce0030104f397321780359d5. actually, i like that version better. thanks for the suggestion.

takezoe:
@ritschwumm Thanks for the fix. I tested and I would like to discuss about form validation.

Currently, baseURL is required when SSH is enabled. However it should be not required when SSH host is entered. I think it might be better that SSH host and port are required when SSH is enabled, and baseURL is optional. How do you think about this?

ritschwumm:
it seemed to me that DefaultGitReceivePack uses CommitLogHook, and that in turn still needs the base url. i'll investigate whether i can get rid of that dependency.

ritschwumm:
not easily doable, i think. the only solution i can imagine is to disable just CommitLogHook when there is no baseUrl, but not git-over-ssh itself.

takezoe:
@ritschwumm Thanks. I understood.

takezoe:
@ritschwumm I added some modifications and merged. Thanks for your pull request!

PR #1085, opened 2016-01-30
properly escape html metacharacters in repository description

ritschwumm:
try entering this description when you create a new project:
```javascript
<script>alert("hallo")</script>
```

### DIFF
--- a/src/main/scala/gitbucket/core/view/helpers.scala
+++ b/src/main/scala/gitbucket/core/view/helpers.scala
@@ -9,7 +9,7 @@ import gitbucket.core.plugin.{RenderRequest, PluginRegistry}
 import gitbucket.core.service.{RepositoryService, RequestCache}
 import gitbucket.core.util.{FileUtil, JGitUtil, StringUtil}
 
-import play.twirl.api.Html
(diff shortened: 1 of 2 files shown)


takezoe:
@ritschwumm Thanks, this is serious bug that should be fixed as soon as possible. But I think more simple fix is enough to avoid XSS like:

```scala
val escaped = HtmlFormat.escape(text)
Html(detectAndRenderLinksRegex.replaceAllIn(escaped, m => s"""<a href="${m.group(0)}">${m.group(0)}</a>"""))
```

ritschwumm:
HtmlFormat.escape(text) returns an Html, not a String.

 i tried with StringUtil.escapeHtml instead, and
```
http://exa<mple.com
```
becomes
```
<a href="http://exa&lt;mple.com">http://exa&lt;mple.com</a>
```
then. that's wrong.

takezoe:
@ritschwumm Right. However it's possible to remove `var` by using `foldLeft` instead of `for` expression. I will try to do it and merge after all your tests are passed.

takezoe:
Merged.

PR #1104, opened 2016-02-13
generic ssh user

ritschwumm:
addresses #971 - let me know what you think.

### DIFF
--- a/src/main/scala/gitbucket/core/ssh/PublicKeyAuthenticator.scala
+++ b/src/main/scala/gitbucket/core/ssh/PublicKeyAuthenticator.scala
@@ -2,22 +2,76 @@ package gitbucket.core.ssh
 
 import java.security.PublicKey
 
+import gitbucket.core.model.SshKey
 import gitbucket.core.service.SshKeyService
 import gitbucket.core.servlet.Database
 import org.apache.sshd.server.auth.pubkey.PublickeyAuthenticator
 import org.apache.sshd.server.session.ServerSession
+import org.apache.sshd.common.session.Session
+import org.slf4j.LoggerFactory
 
-class PublicKeyAuthenticator extends PublickeyAuthenticator with SshKeyService {
+object PublicKeyAuthenticator {
+  // put in the ServerSession here to be read by GitCommand later
+  private val userNameSessionKey = new Session.AttributeKey[String]
 
-  override def authenticate(username: String, key: PublicKey, session: ServerSession): Boolean = {
-    Database() withSession { implicit session =>
-      getPublicKeys(username).exists { sshKey =>
-        SshUtil.str2PublicKey(sshKey.publicKey) match {
-          case Some(publicKey) => key.equals(publicKey)
-          case _ => false
-        }
-      }
+  def putUserName(serverSession:ServerSession, userName:String):Unit =
+    serverSession.setAttribute(userNameSessionKey, userName)
+
+  def getUserName(serverSession:ServerSession):Option[String] =
+    Option(serverSession.getAttribute(userNameSessionKey))
+}
+
+class PublicKeyAuthenticator(genericUser:String) extends PublickeyAuthenticator with SshKeyService {
+  private val logger = LoggerFactory.getLogger(classOf[PublicKeyAuthenticator])
+
+  override def authenticate(username: String, key: PublicKey, session: ServerSession): Boolean =
+    if (username == genericUser) authenticateGenericUser(username, key, session, genericUser)
+    else                         authenticateLoginUser(username, key, session)
+
+  private def authenticateLoginUser(username: String, key: PublicKey, session: ServerSession): Boolean = {
+    val authenticated =
+      Database()
+      .withSession { implicit dbSession => getPublicKeys(username) }
+      .map(_.publicKey)
+      .flatMap(SshUtil.str2PublicKey)
+      .contains(key)
+    if (authenticated) {
+      logger.info(s"authentication as ssh user ${username} succeeded")
+      PublicKeyAuthenticator.putUserName(session, username)
     }
+    else {
+      logger.info(s"authentication as ssh user ${username} failed")
+    }
+    authenticated
   }
 
+  private def authenticateGenericUser(username: String, key: PublicKey, session: ServerSession, genericUser:String): Boolean = {
+    // find all users having the key we got from ssh
+    val possibleUserNames =
+      Database()
+      .withSession { implicit dbSession => getAllKeys() }
+      .filter { sshKey =>
+        Option(sshKey.publicKey)
+        .filter(_.trim.nonEmpty)
+        .flatMap(SshUtil.str2PublicKey)
+        .exists(_ == key)
+      }
+      .map(_.userName)
+      .distinct
+    // determine the user - if different accounts share the same key, tough luck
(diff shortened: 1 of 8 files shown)


nazoking:
great ! I think below

1. getAllKeys is too heavy.
 ( my gitbucket has 89 sshkeys, and select all need 1763 ms (first time, without cache))
 You can index and find public key by it's fingerprint.

2. On authenticateGenericUser, if (username != genericUser), wold you call authenticateLoginUser ?
 for compatibility of user's local repository setting.

ritschwumm:
about your first comment:
- i agree getting all keys and looping over them is at least not elegant
- still, 2 seconds to load a few kilobytes of data is kind of a joke, isn't it?
- how much faster is it after the first lookup? i think two seconds for the very first lookup is ok, if the following ones are faster, which i expect them to be unless you have a lot more keys.
- i don't think i need a fingerprint at all - if SshKeys.publicKey were pre-parsed in the database instead of splitting it in SshUtils.str2PublicKey i could look it up easily, right?

in short: i think the solution i have is good enough for now, but i do agree there is potential for optimization.

about the second comment:
i think's it's a matter of taste. the way it's implemented right now makes it very clear what happens if there already is a user named "git". no special rules, no migration, no nothing. on the other hand, it does put the burden of changing repository urls on the individual user.

nazoking:
> still, 2 seconds to load a few kilobytes of data is kind of a joke, isn't it?

It's not joke.. it is real result, (I try 'SELECT * FROM SSH_KEY;' in my company's gitbucket h2-console ). but now, it 16 ms . Maybe, today is holiday, and , all key is on query cache, and it data is lettle. and, it always on cache if this PR is merged.

> i don't think i need a fingerprint at all - if SshKeys.publicKey were pre-parsed in the database instead of splitting it in SshUtils.str2PublicKey i could look it up easily, right?

do you say about PublicKey.getEncoded or object serialize or decodeBase64(encodedKey) ? i think those data is too large for index.


> if there already is a user named "git".

It is not works well on gitbucket. If you create user nemed "git" , you can't open his page... ( it filterd by [here](https://github.com/gitbucket/gitbucket/blob/5b87efa032d0fa6e1f0c7e0fe88c54301196785b/src/main/webapp/WEB-INF/web.xml#L46) or [here](https://github.com/gitbucket/gitbucket/blob/5b87efa032d0fa6e1f0c7e0fe88c54301196785b/src/main/scala/gitbucket/core/controller/ControllerBase.scala#L60) 
and I think that user named "git" is not exists in all gitbucket system.

ritschwumm:
haha, i could have guessed "git" was not allowed as a username.
how about removing the configuration option completely then, and just unconditionally allow accessing repos as the ssh user "git"? it might confuse users when then ssh urls displayed in the gui change from "user@..." to "git@..." from one release to the next, but i think the reduced complexity would be worth it.

ritschwumm:
so... anyone knows how to implement an inverse of SshUtils.str2PublicKey? more or less every better way of finding the user for a public key depends on having something like that...

nazoking:
>  just unconditionally allow accessing repos as the ssh user "git"?

I think it is good idea. 

> about the index

I don't know how long capacitiy of h2 index, but mysql has limit (gitbucket is trying to support the mysql.). and long varchar index degrades the mysql performance. when user use  4096bit-rsa key, key-string is about 710 chars (if decode it, it is about 530 byte data).

`key.split(" ")(1)` is not sure for index. public-key-string can have line-brake-code (or non-base64 charcters) in string, and gitbucket ignores it now. you can use `Base64.decodeBase64(Constants.encodeASCII(key.split(" ")(1))):Array[Byte]`

why don't you use fingerprint? Fingerprint is not a certainly unique, but it is sufficiently unique, and it is short enough for index.

ritschwumm:
Base64.decodeBase64(Constants.encodeASCII(key.split(" ")(1))) is what SshUtil.str2PublicKey is doing - i'd want the inverse of that

nazoking:
inverse? what type data do you want?

`val decode = Base64.decodeBase64(Constants.encodeASCII(key.split(" ")(1)))` is `Array[Byte]` that is raw-data of public key.

`Base64.encodeBase64(decode)` is `String` without comments ( you need it ?).

`new ByteArrayBuffer(decode).getRawPublicKey` is `PublickKey` it implemented [here](https://github.com/apache/mina-sshd/blob/master/sshd-core/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java#L313)

strict perser of `key` (that user inputed) is maybe [this](https://github.com/apache/mina-sshd/blob/509c871eac5f9dff521891971fb24673135b52c3/sshd-core/src/main/java/org/apache/sshd/common/config/keys/AuthorizedKeyEntry.java#L296) , but it might have no compatibility with gitbucket...

ritschwumm:
simplified the whole thing by removing the settings option:
- as @nazoking mentioned there is no user named "git" the change would cause trouble for
- always have the generic user "git" in ssh urls displayed in the gui reduces possible confusion among users
- both git@... and user@... now work for ssh repository urls, so users don't have to reconfigure their existing (local) repository configuration

ritschwumm:
so... any chance to get this in, or do you require something better, performancewise?

nazoking:
@ritschwumm Reply is sorry now slow .

The thing I worry about is below.

1. User fail to login because public key is ambiguous. When other user use same publick-key.
  In reality, this problem is not problem.
  But when happend, this is hard to understand.  there is no warnings for user (server log only).

2. fetch all public keys and test all keys. My fare is about performance.
  I tested it using real data ( 89 ssh-keys ), But I didn't feel performance problems.
  but...

@takezoe what do you think?

takezoe:
@ritschwumm @nazoking OK, I'm reviewing this pull request now.

src/main/scala/gitbucket/core/ssh/PublicKeyAuthenticator.scala
```diff
@@ -2,22 +2,76 @@ package gitbucket.core.ssh
...
   }
 
+  private def authenticateGenericUser(username: String, key: PublicKey, session: ServerSession, genericUser:String): Boolean = {
+    // find all users having the key we got from ssh
+    val possibleUserNames =
+      Database()
+      .withSession { implicit dbSession => getAllKeys() }
+      .filter { sshKey =>
+        Option(sshKey.publicKey)
+        .filter(_.trim.nonEmpty)
```
takezoe: It's better to move this filter into `getAllKeys()` to reduce unnecessary fetching from database.

takezoe:
@nazoking 
> User fail to login because public key is ambiguous. When other user use same publick-key.

Maybe it should reject a public key which has been already registered by other user in registration.

> fetch all public keys and test all keys. My fare is about performance.

Agree. But I can't find other way. However if performance problem is clarified actually, we can consider some solutions such as in-memory caching.

@ritschwumm In my review, I added a little comment but basically, this request looks good to me. I would like to merge it. Thanks for your contribution.

ritschwumm:
@takezoe the non-blank check is now done in the database; additionally i rebased the PR to 4.0 and checked it's still working

takezoe:
@ritschwumm Great thanks!

takezoe:
@nazoking Please merge! I would like to fix validation for the public key registration after merging.

PR #1105, opened 2016-02-13
make Plugin an abstract class to improve binary compatibility

ritschwumm:
is this everything you wanted to do about #1090, or is there more?

### DIFF
--- a/src/main/scala/gitbucket/core/plugin/Plugin.scala
+++ b/src/main/scala/gitbucket/core/plugin/Plugin.scala
@@ -8,9 +8,9 @@ import gitbucket.core.util.Version
 
 /**
  * Trait for define plugin interface.
- * To provide plugin, put Plugin class which mixed in this trait into the package root.
+ * To provide a plugin, put a Plugin class which extends this class into the package root.
  */
-trait Plugin {
+abstract class Plugin {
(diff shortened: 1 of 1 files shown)


takezoe:
@McFoggy Changing trait breaks binary compatibility certainly in Scala. So using abstract class instead of trait is a good practice to keep binary compatibility.
There are a migration support tool named [migration-manager](https://github.com/typesafehub/migration-manager) originally provided by Typesafe. We can check binary compatibility using this tool.

takezoe:
@ritschwumm Thanks for your work. This modification breaks binary compatibility. So I will merge this request just before next change of `Plugin`.

McFoggy:
@takezoe the usage of clirr just reveal incompatible changes when they come. It does not mean we cannot introduce some when it is expected and fully controled. As you probably already noticed I am a java developer, so I propose the tool I know ;-) But for sure any tooling that does the same is ok to use. Having that in gitbucket tooling would help when enhancements & refactoring are introduced ; or at least it would help to communicate with external consumers (plugin providers mainly).

takezoe:
@McFoggy This request is merged in 3.14 because I would like to add some new extension points to `Plugin`.

PR #1112, opened 2016-02-17
Add margin after octicon

oohira:
Just nitpicking. It's not so smart way because I only added LF char between `<i>` and `<span>` element.

screenshot


oohira:
@McFoggy I agree with you. However, because LF in `.scala.html` is interpreted as space currently, margin size could change depending on whether programmers add LF or not even if we defined some CSS. In ideal way, we should trim whitespace chars in template files automatically and define CSS to add some margin after `<i>` element, but I think it overdone.

Issue #1109, opened 2016-02-15
does logback.xml need to be provided


McFoggy:
Having `logback.xml` provided in `src/main/resources` prevent it to be overidden when launched as a standard java application using `java -jar gitbucket.war`.

- adding a `logback.xml` before gitbucket in the classpath does not work, probably because of the ClassLoading mechanism in Jetty that should first find it in the war before asking to the CCL or AppCL.

  `java -cp .;gitbucket.war JettyLauncher`

- using System property `-Dlogback.configurationFile` does not work (found on some weird SO questions/answers)

The only thing that works (as expressed in [Logback configuration](http://logback.qos.ch/manual/configuration.html#auto_configuration)) is to use a file `logback-test.xml`

As [BasicConfigurator](http://logback.qos.ch/xref/ch/qos/logback/classic/BasicConfigurator.html) is the final fallback mechanism in logback, I am really not sure that gitbucket has to provide a `logback.xml` file doing the exact same and then preventing it to be overidden.

McFoggy:
Just added a [wiki page](https://github.com/gitbucket/gitbucket/wiki/Tracing-%26-logging) to start documenting logging configuration.

Still the question remains for me if it is valuable to keep the `logback.xml` in the war.

lefou:
IMHO, having no easy way to provide / override the logging configuration is something we should try to avoid really hard.

As an idea, we could try to provide a fallback programmatically, if no other configuration was provided to Logback. This might of course bind us to this concrete SLF4J implementation. But as SLF4J ensures, that there is only on implementation, we could try to configure Logback, but silently catch a NoClassDefFoundError / ClassNotFoundException, which would be thrown in the case the user provided a different SLF4J implementation. To work properly, the fallback code must be located in a separate class.

McFoggy:
@lefou today logback + a default loback.xml is packaged within the war. It is bad: yes I agree but it is a fact and it is fully understandable for the simple use case: `java -jar gitbucket.war`

To avoid that for the future, I think we have no other choice than changing the way we produce gitbucket artifacts.
We cannot cover all the needs with a simple all-in-one artifact.

aadrian:
There's also the trick with included files, at least for changing the root logger level:
http://stackoverflow.com/questions/23651466/overriding-logback-configurations#23737143

McFoggy:
@aadrian I proposed already variable substitution in #1122

[...]

PR #1122, opened 2016-02-24
add a chance to override default log level from command line

McFoggy:
First step to better document how to log with gitbucket.

This PR allows for controlling externally the default log level of gitbucket for example by using `java -DlogLevel=DEBUG -jar gitbucket.war`

Also it relates to #1109 it does not fully answer it.
We'll need to document in a better way (with it's own wiki entry probably) how log can be configured for gitbucket.

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -6,23 +6,7 @@
     </encoder>
   </appender>
 
-  <!--
-  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
-    <file>gitbucket.log</file>
-    <encoder>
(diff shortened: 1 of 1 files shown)


takezoe:
@McFoggy I think that output log is too large when root logger is DEBUG level. So this option might be not work effective. It would be better if logger.xml is possible to be replaced at runtime as you mentioned in #1109.

> We'll need to document in a better way (with it's own wiki entry probably) how log can be configured for gitbucket.

Agree.

McFoggy:
Ok let's forget about it. I'll focus instead on documenting how to override logging by providing external configuration.

Issue #1117, opened 2016-02-21
add X-Hub-Signature security to webhooks


McFoggy:
As described in [github documentation](https://developer.github.com/webhooks/#delivery-headers), ones can secure webhooks so that the consumer can trust the webhook.

The required enhancement consist in registring a security token when registering a webhook and to generate an HTTP header when posting the event to the webhook.

[xhub4j](https://github.com/McFoggy/xhub4j) allows to generate easily such headers.

PR #1127, opened 2016-03-01
add X-Hub-Signature security to wekhooks

McFoggy:
as defined in #1117 here is a PR to add X-Hub-Signature handling like defined in [Gitbucket doc](https://developer.github.com/webhooks/#delivery-headers)


src/main/resources/update/3_13.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE WEB_HOOK ADD COLUMN TOKEN VARCHAR(100);
+ALTER TABLE WEB_HOOK_EVENT ADD COLUMN TOKEN VARCHAR(100);
```
nazoking: Why web_hook_event needs token ?

McFoggy: It should not ; I even do not remember why I needed that ; it makes no sense neither to me now.
I'll review the PR tonight.
Thanks for opening my eyes:  less is better... often!

McFoggy:
@nazoking I have simplified the code and had many difficulties in being able to write the simple join between `WebHook` & `WebHookEvent`. That's the reason why I first duplicated the value to the WebHookEvent table ; I didn't needed to do an inner join.
Coming from my pure java/sql/jpa background I can tell you that juming into scala is not easy.

I also rebased on master & squashed the commit

src/main/twirl/gitbucket/core/settings/edithooks.scala.html
```diff
@@ -30,6 +30,16 @@
               }
               <button class="btn btn-default" id="test">Test Hook</button>
             </fieldset>
+            <fieldset class="form-group">
+              <label class="strong">Security Token</label>
+              <div></div>
+              @if(create){
+              <input type="text" name="token" id="token" placeholder="leave blank for no X-Hub-Signature usage" value="@webHook.token" class="form-control" style="display: inline; width: 500px; vertical-align: middle;" />
+              } else {
+              <input type="text" value="@webHook.token" class="form-control" style="display: inline; width: 500px; vertical-align: middle;" disabled />
```
nazoking: Why disabled ? ( github's Secret field is always editable )

McFoggy: In the same way that webhook URL is not editable after creation in gitbucket.

nazoking: oh... I seted disabled to url field, because, 
gitbucket's url field is primary key of web_hook,
And I think it is too complex that unique-check, redirect when change url, validation error...
 ( In github, they created `id`, and use it for primary-key, and it make simple )

But token is not primary key, and I think it sould be editable. However , it may not now .

McFoggy:
@nazoking now token is always editable.

McFoggy:
@takezoe @nazoking once this PR is accepted I'll push another one to add support for content type choosing between `application/json` & `application/x-www-form-urlencoded` like in github

nazoking:
LGTM. @takezoe who can merge it? I do not understand review & merge prosess on gitbucket team.

McFoggy:
@nazoking I fully agree, PR handling, merge, merge policy is something we should have a discussion about for sure with @takezoe

src/main/scala/gitbucket/core/controller/RepositorySettingsController.scala
```diff
@@ -198,7 +199,7 @@ trait RepositorySettingsControllerBase extends ControllerBase {
    * Display the web hook edit page.
    */
   get("/:owner/:repository/settings/hooks/new")(ownerOnly { repository =>
-    val webhook = WebHook(repository.owner, repository.name, "")
+    val webhook = WebHook(repository.owner, repository.name, "", null)
```
takezoe: Set `None` instead of `null`.

takezoe:
@McFoggy I've just added a comment. Please check it.

takezoe:
At first, I would like to avoid database schema changing as much as possible until external database support is completed. I and @shimamoto are progressing in some branches, so we have to pick database changes from the master branch. However, In this request, this is not so serious problem because database changes in this request is small. We can pick them easily.

And before discussion about merge policy, I think we should discuss what features will we make, and declare who is assigned before we start to work. Basically, GitBucket is according to GitHub specification, so we should reject requests which is not belong to this policy.

About merge policy, all organization members can merge bugfix or small request. But in the big request such as:

- Include a lot of source code changes
- Include database changes
- Add new library dependency

It might be better to review by me from point of Scala or design of GitBucket view.

McFoggy:
@takezoe updated with your comments.

For the features, I fully agree with you, gitbucket-core should continue to focus on github functionalities. I benefit from the creation of xhub4j for a professional project to use it also in gitbucket in the same way that it exists in github. My next PR will be to add also content-type selection for webhooks (json / form-urlencoded) exactly like in github.

Regarding my participation in the project and knowing that I am totally new in scala world, I really do not want to merge code PR for now. So I'll personally will continue to answer questions, close issues when relevant, propose PR. Moreover, I'll merge PR only when those ones are related to documentation or real minor fixes.
For @nazoking I think you could let more freedom there.

If you agree perhaps we can continue this discussion by email.

takezoe:
@McFoggy @nazoking 

> If you agree perhaps we can continue this discussion by email.

I would like to open the gitter room for development. We can discuss at there.

PR #1153, opened 2016-03-14
Move to raw Bootstrap3 from GitHub like theme for Bootstrap3

takezoe:


bviktor:
Is there, or will there be an option to still use the GitHub-like interface?

McFoggy:
no there is no way to go back to that interface.
We had to change because of a github demand (expressed by gthub layers) to move away from their UI that is copyrighted.

bviktor:
Oh wow, that's quite lame of them. Thanks for the insight!

PR #1158, opened 2016-04-01
Fix layouts for Issue pages.

ledyba:
In this PR, I'm fixing a layout problem in issues pages.

It looks that we have to use "row-fluid" class instead of "row" class, because we do not use fixed-size layout from 3.13.
FYI:
(English) http://stackoverflow.com/questions/25682804/row-fluid-vs-row-in-twitter-bootstrap
(Japanese) http://greenapple-room.com/conc/user/TwitterBootstrap/bootstrap_02.html

Could you take a look?

## Before
before
Labels/Milestone, ... are in bottom of comments.

## After
after
Labels/Milestone, ... are in right of comments.


takezoe:
@ledyba Thanks for your pull request!

We will make large update Issues and Pull requests UI in the next release.
Right sidebar might be abolished.

PR #1236, opened 2016-07-06
Fix plugin versions in installed plugins page

mrkm4ntr:
Solidbase migration requires ascending order of versions.
I fix this issue with a workaround because many existing plugins have descending order.

### DIFF
--- a/src/main/scala/gitbucket/core/plugin/PluginRegistory.scala
+++ b/src/main/scala/gitbucket/core/plugin/PluginRegistory.scala
@@ -203,7 +203,7 @@ object PluginRegistry {
           instance.addPlugin(PluginInfo(
             pluginId    = plugin.pluginId,
             pluginName  = plugin.pluginName,
-            version     = plugin.versions.head.getVersion,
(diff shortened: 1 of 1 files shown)


takezoe:
@mrkm4ntr Certainly, Solidbase behavior might confuse. But essentially, existing plugins should move to the new plugin version management system based on Solidbase.

In addition, Solidbase does not force that version number is `x.x.x` and numeric. For example, it's allowing `x.x.x-SNAPSHOT` as the version number. So this workaround has possibility of causing error.

McFoggy:
what about moving to a [semver](http://semver.org) compatible implementation of `Version`? doing so versions, comparison & ranges could follow a standard.

mrkm4ntr:
@takezoe Fair enough. I changed this commit to what should be.
@McFoggy Thank you for your proposal. But I decided not care about existing plugins, because this effect is not critical.

takezoe:
@McFoggy Thanks for suggestion. I will consider implementing it in Solidbase.

PR #1237, opened 2016-07-08
Fix some layout CSS

rlazoti:
This PR fixes two issues in the Gitbucket's CSS.

* The first one fixes the CSS for some divs (comments) floating outside the content layout.

  Example:
image

  After the fix:
image

* The second fixes the CSS of the check-conflict box.

  Example:
image

  After the fix:
image

### DIFF
--- a/src/main/twirl/gitbucket/core/menu.scala.html
+++ b/src/main/twirl/gitbucket/core/menu.scala.html
@@ -58,7 +58,7 @@
     </div>
   </div>
   <div class="content-wrapper">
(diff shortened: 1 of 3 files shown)


alexkuznetsov:
Maybe it would be better to create a CSS classes for all inline css in tags?

takezoe:
@rlazoti Thanks for your request! I fixed these broken layouts by other way. See diff: cf9a55d896b496b247763af8d82393b8750a4f37

PR #1249, opened 2016-07-25
fix: can't resolve the git repository path provided by the plugin

shiena:
If the contextPath is not equals to `/`, gitRepositoryPath contains
the contextPath + `/git`.  Therefore, gitbucket can not resolve
the git repository path provided by the plugin.

For example, you can not clone the snippets repository of gist-plugin.

To fix this, also deletes contextPath from requestURI.

### DIFF
--- a/src/main/scala/gitbucket/core/util/Implicits.scala
+++ b/src/main/scala/gitbucket/core/util/Implicits.scala
@@ -73,7 +73,7 @@ object Implicits {
 
     def hasAttribute(name: String): Boolean = request.getAttribute(name) != null
 
-    def gitRepositoryPath: String = request.getRequestURI.replaceFirst("^/git/", "/")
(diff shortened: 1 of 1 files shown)


src/main/scala/gitbucket/core/util/Implicits.scala
```diff
@@ -73,7 +73,7 @@ object Implicits {
 
     def hasAttribute(name: String): Boolean = request.getAttribute(name) != null
 
-    def gitRepositoryPath: String = request.getRequestURI.replaceFirst("^/git/", "/")
+    def gitRepositoryPath: String = request.getRequestURI.replaceFirst("^" + request.getContextPath + "/git/", "/")
```
nazoking: contextPath can contain some special chars ( for example, dot). you should escape contextPath. maybe you can use `java.util.regex.Pattern.quote`.

shiena:
@nazoking Thank you for the advice. I was modified contextPath to the literal pattern.

Issue #447, opened 2014-07-28
Add a local file to repository


chenjb1975:
Currently, I can add a brant new file to repository. And edit it online.
But sometime I need add a local file to repository via web interface.

PR #1252, opened 2016-07-27
Enable file upload into repository

devagorilla:
I am proposed to add file upload functionality to repo, similar to what Github currently has

This is just a prototype, but I would like to get some feedback from you guys. Currently it only supports text based files, but I would like to include all file type

### DIFF
--- a/src/main/scala/gitbucket/core/api/UploadFiles.scala
+++ b/src/main/scala/gitbucket/core/api/UploadFiles.scala
@@ -1,0 +1,12 @@
+package gitbucket.core.api
+
+
+case class UploadFiles(branch: String, path: String, fileIds : Map[String,String], message: String)
+{
+  def isValid: Boolean = {
+
+    fileIds.size > 0
+  }
+
+
+}
--- a/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
+++ b/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
@@ -1,7 +1,8 @@
 package gitbucket.core.controller
 
-import javax.servlet.http.{HttpServletResponse, HttpServletRequest}
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 
+import gitbucket.core.api.UploadFiles
 import gitbucket.core.plugin.PluginRegistry
 import gitbucket.core.repo.html
 import gitbucket.core.helper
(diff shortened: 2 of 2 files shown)


src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
```diff
@@ -526,6 +526,118 @@ trait RepositoryViewerControllerBase extends ControllerBase {
...
+    }
+  })
+
+  /*
+Roy Li modification
+ */
+
+  case class CommitFile(fileId: String, name: String, fileBytes: Array[Byte])
+
+  private def commitFiles(repository: RepositoryService.RepositoryInfo,
```
takezoe: Existing `RepositoryViewerController.commitFile()` might be possible to be replaced with this method.

src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
```diff
@@ -526,6 +526,118 @@ trait RepositoryViewerControllerBase extends ControllerBase {
...
+    defining(repository.owner, repository.name){ case (owner, name) =>
+      (for {
+        data <- extractFromJsonBody[UploadFiles] if data.isValid
+      } yield {
+        Directory.getAttachedDir(owner, name) match {
+          case dir if (dir.exists && dir.isDirectory) =>
+            val _commitFiles = data.fileIds.map { case (fileName, id) =>
+              dir.listFiles.find(_.getName.startsWith(id + ".")).map { file =>
+                val s = scala.io.Source.fromFile(file) // Codec ????
+              val byteArray = s.map(_.toByte).toArray
```
takezoe: You can use `IOUtils` instead of `scala.io.Source` to get bytes like this:
https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/controller/FileUploadController.scala#L66

src/main/scala/gitbucket/core/api/UploadFiles.scala
```diff
@@ -0,0 +1,12 @@
+package gitbucket.core.api
+
+
+case class UploadFiles(branch: String, path: String, fileIds : Map[String,String], message: String)
```
takezoe: Why this class is defined in `gitbucket.core.api` package?

takezoe:
@devagorilla Great attempt! I added some comments to your code, and I'm looking forward to a complete request to make file uploading on the repository viewer.

src/main/scala/gitbucket/core/api/UploadFiles.scala
```diff
@@ -0,0 +1,12 @@
...
+{
+  def isValid: Boolean = {
```
takezoe: lazy val might be better in this case.

aadrian:
https://github.com/gitbucket/gitbucket/issues/447

takezoe:
Resolved conflicts in #1605

Issue #935, opened 2015-10-11
Deploy gitbucket-assembly.jar into Maven central or CDN


takezoe:

ritschwumm:
see https://github.com/gitbucket/gitbucket/pull/1065 for some more details

McFoggy:
any info on that? amatareas.sourceforge.jp is really slow from europe.

takezoe:
@McFoggy Requested `io.github.gitbucket` group id to sonatype with @tanacasino and your account I found in sonatype JIRA: https://issues.sonatype.org/browse/OSSRH-20869

I would like to publish other libraries of GitBucket organization such as markedj, solidbase and scalatra-forms to Maven central repository, not only `gitbucket-assembly.jar`.

McFoggy:
@takezoe for sonatype once you have register a domain then you can pubish as many artifacts as you want within this domain (or any subdomain). I did exactly that for my projects under [oss.brouillard.fr](http://oss.brouillard.fr/). But I could do it because I own the domain `brouillard.fr`. 
It is perhaps time for gitbucket organization to collect some money at least to pay for a domain. If we host only static content, github pages can be sufficient.

McFoggy:
Without a domain, once you have the green light for `io.github.gitbucket`, then you will be able to deploy also any other artifact if an only if its groupid is `io.github.gitbucket` or any child of it `io.github.gitbucket.libraries`, `io.github.gitbucket.markedj`.

takezoe:
@McFoggy I know. I already have some other group ids in sonatype.

I was looking for domain for GitBucket few years ago. However I couldn't get suitable domain such as `gitbucket.com`. Unfortunately, they had been acquired to other people.

Anyway, I will publish all jars to the Maven central repository after my request is accepted. Please wait for a moment.

McFoggy:
@takezoe any chance to finally have the gitbucket artifacts deployed to central?

PR #1253, opened 2016-07-31
(refs #935) Update project configuration for deploying artifacts to Maven central

takezoe:
This update makes possible to publishing artifacts to Maven central for plugins.

One problem remained is both of jar and war are published. A war file shouldn't be published. Currently, I have to delete a war file and its hash on Nexus after publish artifacts to Sonatype.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 val Organization = "io.github.gitbucket"
 val Name = "gitbucket"
-val GitBucketVersion = "4.4.0"
+val GitBucketVersion = "4.3.0"
 val ScalatraVersion = "2.4.1"
 val JettyVersion = "9.3.9.v20160517"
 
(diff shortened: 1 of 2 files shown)


xuwei-k:
> One problem remained is both of jar and war are published. A war file shouldn't be published. Currently, I have to delete a war file and its hash on Nexus after publish artifacts to Sonatype.

please try to this

```scala
artifact in Keys.`package` := Artifact(moduleName.value)
```

xuwei-k:
default setting in sbt
- https://github.com/sbt/sbt/blob/v0.13.12/main/src/main/scala/sbt/Defaults.scala#L614

in xsbt-web-plugin
- https://github.com/earldouglas/xsbt-web-plugin/blob/2.1.0/src/main/scala/com/earldouglas/xwp/WarPlugin.scala#L12


`artifact in Keys.`package` := Artifact(moduleName.value)` means **Re**-override the `artifact` value.

takezoe:
@xuwei-k It works! Thanks a lot!!

PR #1269, opened 2016-08-13
[WIP] Move to Slick3 using blocking-slick

takezoe:
This pull request is an experiment of moving to Slick3 using blocking-slick library.
https://github.com/takezoe/blocking-slick


hrj:
Coding is a little easier with a blocking API; I wish Scala had support for `await` keyword so that we could consume async APIs in blocking style.

What's the benefit *you* see in moving to blocking API?

takezoe:
@hrj Once, we tried to move to Slick3 by blocking `Future` using `Await.result`. But it was too hard, so we gave up to move on this way:
https://github.com/gitbucket/gitbucket/compare/slick-3.0?expand=1

In addition, blocking `Future` requires `ExecutionContext` (similar to thread pool), it brings unnecessary complexity for synchronous execution.

takezoe:
Merged with `scala-2.12` branch. See #1381.

PR #1273, opened 2016-08-22
Set default value of parameter in Get-Contents API

kw-udon:
If parameter `ref` is not given in get-contents API, repository’s default branch will be used.
https://developer.github.com/v3/repos/contents/#get-contents


takezoe:
@kw-udon GitHub API specification is saying:
> Default: the repository’s default branch (usually master)
So former implementation was wrong. Thanks a lot!

PR #1321, opened 2016-10-10
Improve GitHub compatibility for Jenkins

int128:
I am using [Jenkins Multibranch Pipeline](https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Multibranch+Plugin) and found it does not work with GitBucket. It assumes Git client can access to a repository via GitHub compatible URL. Also it assumes Users API serves not only a user but also an organization (this is current behavior of GitHub API but not documented on [specification](https://developer.github.com/v3/users/)).

This pull request contains ~~2 commits~~ a commit to improve GitHub compatibility as follows.

1. Add redirection to allow Git client to access a repository by GitHub compatible URL, e.g. redirection from `/int128/gitbucket.git/` to `/git/int128/gitbucket.git/`.
2. ~~Change Users API serves also organization as GitHub API, e.g. `/api/v3/users/gitbucket` serves the organization `gitbucket`.~~ This is not needed on the latest Multibranch Pipeline.

I have tested these features on both my Mac and EC2 server terminated with ELB SSL.

Additionally, I am happy to contribute a Wiki page how to use Jenkins Multibranch Pipeline and GitBucket. For more details, please see [my blog post](http://int128.hatenablog.com/entry/2016/10/06/224444) (Japanese).

Thanks for the great work.

### DIFF
--- a/src/main/scala/ScalatraBootstrap.scala
+++ b/src/main/scala/ScalatraBootstrap.scala
@@ -1,12 +1,12 @@
 
-import gitbucket.core.controller._
-import gitbucket.core.plugin.PluginRegistry
-import gitbucket.core.servlet.{ApiAuthenticationFilter, Database, GitAuthenticationFilter, TransactionFilter}
-import gitbucket.core.util.Directory
 import java.util.EnumSet
 import javax.servlet._
 
+import gitbucket.core.controller._
+import gitbucket.core.plugin.PluginRegistry
 import gitbucket.core.service.SystemSettingsService
+import gitbucket.core.servlet._
+import gitbucket.core.util.Directory
 import org.scalatra._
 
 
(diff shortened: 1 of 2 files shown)


McFoggy:
Are the 2 commits really related? Isn't the first one enough for the github plugin?  
You are mentioning the multibranch plugin but the multibranch plugin does not interfere here with the git url ; only the github plugin expects some convention/special url for the repository ; or am I wrong?

int128:
I checked the latest version of Multibranch Pipeline and it works fine without the second commit. I will remove the second commit that may break confirmation to the API specification.

@McFoggy The redirection (first commit) works on both GitHub Plugin and Multibranch Pipeline because Pipeline internally uses GitHub Plugin.

takezoe:
@int128 Thanks for your fix 👍

McFoggy:
> @int128 _because Pipeline internally uses GitHub Plugin_

Are you sure of that? I do not think so. 

image

But for sure, if you want to ease Pipeline trigerring and especially be able to build PR, then you need the github plugin. 

At my work I am using the multibranch without the Github plugin, I am triggering the build via the "build with token" feature directly on the multibranch project having a fake webhook in gitbucket pointing to the URL of this token (so no github webhook specific rest endpoint listebing for the webhook). 
This setup works well for all branches but does not work for any PR, coming from the repository itself or from forks. For that later use case, you indeed need the github plugin and the compatible URL you provided.

McFoggy:
@takezoe I agree this PR (first commit) is usefull to be more compatible with github jenkins plugin. It serves the same purpose than the nginx rewrite rules I have posted a while ago on the gitbucket_dev gitter chat room. So yes I agree we should integrate it.

int128:
@McFoggy I think so because both GitHub plugin and the multibranch with GitHub (specify GitHub as a branch source; maybe using GitHub plugin?) do not work without this fix. I am not sure about implementation of plugins and it might be wrong.

As you mentioned, the multibranch with "build with token" does not work for PRs. The multibranch with GitHub is needed for projects using PR based flow such as GitHub flow. This contribution may be valuable for Jenkins users using PR based flow.

Issue #1318, opened 2016-10-07
Commits not appear in front-page since 4.5.0


mariodebian:
## Issue

**Impacted version**: 4.5.0
**Deployment mode**: _apache2 mod_proxy_

**Problem description**:
- _Since upgrade to 4.5.0 commits don't appear in front page (AKA News Feed) only issues_

yanma:
I'm curious about this phenomenon. Maybe I faced the same issue as your report.
As I commented in the issue #993, the news feed doesn't appear when I push the commit over ssh.
Could you describe your problem in detail?

mariodebian:
Commits and pull-requests made with WebUI shows ok, but commits sent with ssh (don't test http) not appear in news feed.

McFoggy:
yesterday evening before the resolution of #1319 and while trying to reproduce it on the heroku demo site I tried to do some commits from the UI and didn't noticed neither that commits were referenced in issues ; perhaps it is related.

takezoe:
Fixed in #1336

PR #1336, opened 2016-10-30
(refs #1318) make record***Activity via ssh works again

yanma:
### Impacted version

4.5.0 and 4.6.0
### Problem

In commit 2ca20af, "auto commit" feature is disabled in TransactionFilter.scala. Because it prevents gitbucket from recognizing ssh push from the client, it affects several issues as below.
- Issue #1318 : Commits does not appear in front-page.
- Pull-request is not updated (It is reported [here](https://gitter.im/gitbucket/gitbucket_ja?at=58003f8c457ae29b71d9cdc2)).

I think it can affect any other phenomenons which gitbucket needs record***Activity.
Especially, pull-request is key feature of the gitbucket.
So, I think this is critical issue to be resolved as soon as possible. I hope 4.6.1 is released...
### Solution

My solution is very simple.
Because I'm not sure why it is set as false, I made "auto commit" feature configurable.
But I think it shall be set as true. So, I changed its default value true.

I confirmed all tests passed and the phenomenon diapered.
I hope it works well. Thank you!

### DIFF
--- a/src/main/scala/gitbucket/core/ssh/GitCommand.scala
+++ b/src/main/scala/gitbucket/core/ssh/GitCommand.scala
@@ -36,7 +36,7 @@ abstract class GitCommand extends Command with SessionAware {
     override def run(): Unit = {
       authUser match {
         case Some(authUser) =>
-          Database() withSession { implicit session =>
+          Database() withTransaction { implicit session =>
             try {
               runTask(authUser)
               callback.onExit(0)


takezoe:
@yanma It should enable transaction instead of enabling auto commit. I think it's possible to enable transaction for SSH access by replacing `withSession` with `withTransaction` in the following line:
https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/ssh/GitCommand.scala#L39

yanma:
@takezoe Thank you for your advice. I tried it and confirmed it can work. I updated my pull request.

takezoe:
@yanma Thanks for your work. Could you rebase all commits in this request as last one commit?

yanma:
@takezoe I rebased my pull request. Please check it.

takezoe:
@yanma Great! Thanks a lot.

PR #1342, opened 2016-11-03
Added per-repository issue template feature.

imeszaros:
This feature allows repository administrators to enter a default issue text on the repository options form. On new issue creation this default text is presented in the issue description textarea.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 val Organization = "io.github.gitbucket"
 val Name = "gitbucket"
-val GitBucketVersion = "4.6.0"
+val GitBucketVersion = "4.7.0"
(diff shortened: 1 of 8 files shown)


McFoggy:
wouldn't such a feature be better if compliant with existing github behavior?
Instead of having a dedicated configuration per repository we could lookup the template under:
-  `.github/ISSUE_TEMPLATE.md` to be github compliant
-  `.gitbucket/ISSUE_TEMPLATE.md` to have our own default place to lookup.

imeszaros:
@McFoggy I agree with you. Unfortunately i'm not a GitHub guru, but i missed this feature. I'm looking forward to change this as you proposed.

PR #1359, opened 2016-11-29
Updated admin lte to 2.3.8

hrj:
I found a problem with alignment of labels and wanted to ensure that the latest AdminLTE has the problem. It [does](https://github.com/almasaeed2010/AdminLTE/issues/1260).

As a side-effect I have updated the AdminLTE version.


takezoe:
@hrj Thanks for your request. It needed to fix JavaScript importing too, so I fixed in d3cdc5d5fc6e3314de44637f3abc5c909bf921b3.

Issue #1348, opened 2016-11-11
Keep comment on source code even if new commits are pushed


takezoe:
**Impacted version**: 4.6

**Problem description**:
Comment on source code is disabled if new commits are pushed in the current version of GitBucket. However If new commits don't updated commented line, it's good to keep comment.

PR #1373, opened 2016-12-11
(refs #1348)Keep pull request comment if new commits are pushed

takezoe:


takezoe:
TODO: Comments on the old line are removed.

Issue #1057, opened 2016-01-20
Gitbucket webhook error with diff.algorithm=patience


hanxue:
Hello,

When I test Jenkins Webhook, I don't get any response

![Gitbucket webhook fail](https://cloud.githubusercontent.com/assets/760008/12443423/bacdb6e6-bf92-11e5-8b25-12230854767e.png)

From the HTTP response, I see this

`java.lang.IllegalArgumentException: Invalid value: diff.algorithm=patience`

Full Java stack trace at https://gist.github.com/hanxue/c9917be96c5d010b8cc5

After removing the `git.config` for `diff.algorithm=patience`, webhook works again. 
### Request
1. Can you allow any valid `diff.algorithm` settings?
2. Display an error message and the stack trace when WebHook test fails, rather than fail silently

Especially on (2), I spent 2 days twiddling with Jenkins and Maven, because I set up webhook with gitbucket in the past and it worked like a charm :smile:

PR #1402, opened 2017-01-03
(refs #1075)show alert on webhook ajax error.

nazoking:
alert
image


nazoking:
oh.. original issue id is #1057

PR #1412, opened 2017-01-11
Add support for SMTP STARTTLS

### LINKED ISSUE #1410: Cannot send emails with STARTTLS
Hi,

Office 365 SMTP requires STARTTLS instead of SSL for sending emails. There's no option to enable that in the GitBucket admin panel.

For Jenkins the workaround was adding

~~~
JAVA_OPTS="-Dmail.smtp.starttls.enable=true"
~~~

But here adding that to the Tomcat (in which I run GitBucket) config won't fix the problem. There's no related information in the logs. All the website says is:

> [Error] org.apache.commons.mail.EmailException: sending the email to the following server failed : smtp.office365.com:587

I can't find any related log whatsoever. Tried to run GitBucket in standalone mode with
~~~
java -Dlogback.configurationFile=/opt/gitbucket/config/logback-settings.xml -jar gitbucket.war
~~~

But the mentioned logfiles under `/opt/gitbucket/log` are not created and STDOUT doesn't produce any useful information either.

I tried to patch Notifier.scala with:

~~~
email.setStartTLSEnabled(true)
email.setStartTLSRequired(true)
~~~

But it didn't fix the problem. Additionally, when I try to run my own .war file in standalone mode, it says

~~~
Error: Could not find or load main class JettyLauncher
~~~

So I'm pretty much out of ideas. Did anyone have luck figuring this out?

bviktor:
Fixes #1410

STARTTLS is mandatory for certain SMTP providers, such as Office 365. This PR adds that option, tested and veified to be working with our subscription.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala
+++ b/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala
@@ -41,6 +41,7 @@ trait SystemSettingsControllerBase extends AccountManagementControllerBase {
         "user"                     -> trim(label("SMTP User", optional(text()))),
         "password"                 -> trim(label("SMTP Password", optional(text()))),
         "ssl"                      -> trim(label("Enable SSL", optional(boolean()))),
+        "starttls"                 -> trim(label("Enable STARTTLS", optional(boolean()))),
         "fromAddress"              -> trim(label("FROM Address", optional(text()))),
         "fromName"                 -> trim(label("FROM Name", optional(text())))
     )(Smtp.apply)),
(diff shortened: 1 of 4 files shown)


McFoggy:
Just as a side question, does this STARTTLS flag is linked (or should be linked) to the SSL one?

bviktor:
McFoggy, definitely not. For instance, Office 365 requires SSL to be disabled and STARTTLS to be enabled.

McFoggy:
LGTM

takezoe:
@bviktor Merged. Thanks for your request!

bviktor:
Thanks a lot!

PR #1413, opened 2017-01-11
Fix erroneous label for definition, assumably caused by copy-paste

bviktor:
Title says it all.


takezoe:
@bviktor Thanks for fix :-)

bviktor:
Thank you!

PR #1419, opened 2017-01-12
Issue and PR template per repository

tomoki1207:

Related #1342.
When Issue/PR creating, set initial content via template file in project root or `.gitbucket` folder. (compliant with [GitHub behavior](https://github.com/blog/2111-issue-and-pull-request-templates))

### DIFF
--- a/src/main/scala/gitbucket/core/service/RepositoryService.scala
+++ b/src/main/scala/gitbucket/core/service/RepositoryService.scala
@@ -3,7 +3,14 @@ package gitbucket.core.service
 import gitbucket.core.controller.Context
 import gitbucket.core.model.{Collaborator, Repository, RepositoryOptions, Account, Role}
 import gitbucket.core.model.Profile._
+import gitbucket.core.plugin.PluginRegistry
 import gitbucket.core.util.JGitUtil
+import gitbucket.core.util.JGitUtil.FileInfo
+import gitbucket.core.util.ControlUtil._
+import gitbucket.core.util.Directory
+import gitbucket.core.util.FileUtil
+import gitbucket.core.util.StringUtil
+import org.eclipse.jgit.api.Git
 import profile.simple._
 
 trait RepositoryService { self: AccountService =>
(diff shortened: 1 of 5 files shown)


src/main/scala/gitbucket/core/service/RepositoryService.scala
```diff
@@ -416,6 +423,30 @@ trait RepositoryService { self: AccountService =>
...
+    *
+    * @param repository the repository information
+    * @param fileBaseName the file basename of template
+    * @return The content of template if the repository has it, otherwise empty string.
+   */
+  def getContentTemplate(repository: RepositoryInfo, fileBaseName: String)(implicit s: Session): String = {
+    val preferred = PluginRegistry().renderableExtensions.map(extension => s"${fileBaseName}.${extension}")
+
+    def choiceTemplate(files: List[FileInfo]): Option[FileInfo] =
+      files.find(f => preferred.contains(f.name)).orElse(files.find(f => f.name.startsWith(fileBaseName)))
```
McFoggy: why not filtering with exact naming (with or without extension)?
here I assume we can retrieve file names like : `ISSUE` or ISSUE_TEMPLATE_THAT_SHOULD_NOT_BE_SELECTED

tomoki1207: For retrieve template file regardless its extensions, I implemented like that.
But as you said, it is bug what file named start with specific word (like ISSUE_TEMPLATE_XXXXX) is treated as template also.
So, I'll fix to check if the full basename of file is equal to specific word when `.md` and `.markdown` are not found.

src/main/scala/gitbucket/core/service/RepositoryService.scala
```diff
@@ -434,7 +434,7 @@ trait RepositoryService { self: AccountService =>
     val preferred = PluginRegistry().renderableExtensions.map(extension => s"${fileBaseName}.${extension}")
 
     def choiceTemplate(files: List[FileInfo]): Option[FileInfo] =
-      files.find(f => preferred.contains(f.name)).orElse(files.find(f => f.name.startsWith(fileBaseName)))
+      files.find(f => preferred.contains(f.name)).orElse(files.find(f => fileBaseName == FileUtil.trimExtension(f.name)))
```
McFoggy: can't we simplify with something like _(untested)_

`files.find(f => preferred == f.name).orElse(files.find => fileBaseName == f.name)`

with file extension removal we could select `ISSUE_TEMPLATE.TXT` instead of `ISSUE_TEMPLATE`
I would tend to KISS principle.

tomoki1207: `preferred` is list of `ISSUE_TEMPALTE.md` and `ISSUE_TEMPLATE.markdown`.
And I think `ISSUE_TEMPLATE.TXT` should be treated as template if `.md` not found.
(But I'm not considering that case of `ISSUE_TEMPLATE` and `ISSUE_TEMPLATE.TXT` are exist both)

McFoggy: sorry by just commenting the diffs I missed the real type.

I just tested how it really behaves on github in [McFoggy/jira-example-project](https://github.com/McFoggy/jira-example-project).

According to my few tests github trys to load files in the following order:
- FILE
- FILE.[whatever extension] where extensions are order by lexicographic order

Thus regarding your PR I would just order the preferred list to pickup the first defined renderer, and then try to find a file which name starts with fileBaseName.`  (notice the dot)

takezoe: Although markup renderer can be added by plugin, the content of issues and pull request are rendered as markdown certainly. So it doesn't need to look `PluginRegistry().renderableExtensions` at here. I think that only the following filenames are sufficient for the target.

- `ISSUE_TEMPALTE`
- `ISSUE_TEMPALTE.md`
- `ISSUE_TEMPALTE.markdown`

tomoki1207: @McFoggy
Thank you for testing. I understood the file without extension takes precedence.

@takezoe 
OK, I'll fix that way.
Should we set priorities for these template targets? Also, are template names case sensitive?

takezoe: @tomoki1207 I don't not mind about details. You can do as you prefer ;-)
If you can't determine the specification, it's good to reference GitHub behavior.

tomoki1207: According to [this](https://help.github.com/articles/creating-an-issue-template-for-your-repository/),  template files are not case sensitive.
And files without extensions seem to have higher priority than files with extensions(as @McFoggy said).
So, I'll fix as above. Please just a moment.

tomoki1207:
I committed and rebased.
Should I squash all my commits?

takezoe:
@tomoki1207 Yes, please.

takezoe:
@tomoki1207 Merged. Thanks for your work!

PR #1420, opened 2017-01-12
bump embedded-mysql for windows support

nazoking:
Older versions fail because they do not support testing on windows.

### DIFF
--- a/src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
+++ b/src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
@@ -33,6 +33,7 @@ class GitBucketCoreModuleSpec extends FunSuite {
       .withPort(3306)
       .withUser("sa", "sa")
       .withCharset(Charset.UTF8)
+      .withServerVariable("log_syslog", 0)
       .build()
 
     val mysqld = anEmbeddedMysql(config)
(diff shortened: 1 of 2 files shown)


src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
```diff
@@ -33,6 +33,7 @@ class GitBucketCoreModuleSpec extends FunSuite {
       .withPort(3306)
       .withUser("sa", "sa")
       .withCharset(Charset.UTF8)
+      .withServerVariable("log_syslog", 0)
```
nazoking: Normal Windows users do not have permission to access the event log and the test will fail.

src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
```diff
@@ -41,7 +42,7 @@ class GitBucketCoreModuleSpec extends FunSuite {
 
     try {
       new Solidbase().migrate(
-        DriverManager.getConnection("jdbc:mysql://localhost:3306/gitbucket", "sa", "sa"),
+        DriverManager.getConnection("jdbc:mysql://localhost:3306/gitbucket?useSSL=false", "sa", "sa"),
```
nazoking: > WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.

PR #1428, opened 2017-01-17
use value class

xuwei-k:


McFoggy:
Not being a scala dev, what is the improvement?
I do not see any visible direct usage in the commit and it is always time to learn from other commits.

takezoe:
@McFoggy The value class is a feature to avoid object allocation, although it has some constraints:
http://docs.scala-lang.org/overviews/core/value-classes.html

Implicit classes that is not a value class make a new instance for implicit type conversion. We can avoid this overhead by using the value class.

McFoggy:
Ok I see now why I could not see any direct use. Thanks a lot.

Issue #503, opened 2014-09-26
Failed login gives no error message


ziggystar:
No hint on whether user does not exist, is suspended or password is wrong. This would be **very** helpful.

takezoe:
Fixed in #1434 and #499

PR #1434, opened 2017-01-25
Show login failed message

tomoki1207:

Related to #503. (but, I think it will not be solved completely, it won't be closed by this PR)

Show login failed message without detail reason.
When above, remain username and password user entered.
sign in

### DIFF
--- a/src/main/scala/gitbucket/core/controller/IndexController.scala
+++ b/src/main/scala/gitbucket/core/controller/IndexController.scala
@@ -49,13 +49,18 @@ trait IndexControllerBase extends ControllerBase {
     if(redirect.isDefined && redirect.get.startsWith("/")){
       flash += Keys.Flash.Redirect -> redirect.get
     }
-    gitbucket.core.html.signin()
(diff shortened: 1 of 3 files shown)


src/main/scala/gitbucket/core/controller/IndexController.scala
```diff
@@ -55,7 +55,7 @@ trait IndexControllerBase extends ControllerBase {
   post("/signin", signinForm){ form =>
     authenticate(context.settings, form.userName, form.password) match {
       case Some(account) => signin(account)
-      case None          => redirect("/signin")
+      case None          => gitbucket.core.html.signin(form.userName, form.password)
```
takezoe: In this case, we should use flash attribute. Reference:
https://github.com/gitbucket/gitbucket/blob/7c484297d7f0343c78b64c11dc668de613cc9cb7/src/main/scala/gitbucket/core/controller/SystemSettingsController.scala#L150-L168

tomoki1207:
Done! Anything else?

takezoe:
@tomoki1207 Great! Thanks for your quick fix.

Issue #777, opened 2015-06-05
Set http header for browser cache


nazoking:
Please set http header for browser cache control. For examples ETag , Last-Modified, Expires.

Especially, under /assets, and avater.

( I looked for an effective method in scalatra and jetty , but I did not know .. ;)

haarp:
Yes, please do. I notice that avatars, for example, are always reloaded on every page refresh. Something is disabling the browser cache on Gitbucket.

haarp:
If you're running an Apache reverse proxy in front of Gitbucket, you can work around this bug with:

        <LocationMatch "_avatar$">
                Header set Cache-Control "max-age=86400, public"
        </LocationMatch>

It really only seems to affect avatars, all other resources are cache properly by the browser (HTTP 304)

kounoike:
In current version of GitBucket (4.10 + some patch) and chrome, /assets/* are cached. But /plugin-assets not cached.

![cache](https://cloud.githubusercontent.com/assets/6997928/23854817/7288b38e-0836-11e7-84f2-9c0cf5349748.png)

I read HTTP session by Fiddler, /assets/* has Last-Modified header and Date header. /plugin-assets/* has Date header only.
By #1489, some CSS move from /assets to /plugin-assets. It makes bad feel to user's experience?

kounoike:
|Browser|/assets|/plugin-assets|_avatar|
|---------|-------|---------------|--------|
|Chrome|cached|not cached|not cached|
|firefox|cached|not cached|not cached|
|IE 11|not cached|not cached|not cached|

/plugin-assets served by PluginAssetsServlet, then it can easy to add some HTTP headers.
_avatar served by AccountController, I don't know how to add HTTP headers. Because http://www.scalatra.org/2.4/guides/http/requests-and-responses.html is down now.

kounoike:
Oh... my IE's "常にサーバから更新する" (always reload from servers) setting is checked.
I removed this check, /assets/* are cached.

And I add `resp.setHeader("Cache-Control", "max-age=86400")` to PluginAssetsServlet, /plugin-assets also cached.

takezoe:
Fixed in #1491

PR #1489, opened 2017-03-12
Add Theme Plugin support

kounoike:
This PR introduce "Theme Plugin" support. discoused in #1279.

Each GitBucket instance has some "theme"s. And each theme has some "skin"s.
ex.

- [AdminLTE theme](https://github.com/kounoike/gitbucket-theme-AdminLTE-plugin) is theme of AdminLTE theme.
    - skin-blue
    - skin-blue-light
    - skin-black
    - skin-black-light, etc.
- [AdminLTE pink theme](https://github.com/kounoike/gitbucket-theme-AdminLTE-pink-plugin) is demonstrate version theme.
    - skin-pink

In System settings, You can chose "skin name" with theme/skin pair.

140328-system settings - google chrome

first option is adminLTE-pink theme's skin. others are AdminLTE theme's skin.

theme/skin setting is saved into GITBUCKET_HOME/gitbucket.conf.

Every page of GitBucket load only selected theme's CSS/JS.

If you didn't have no theme-plugins, fallback to previous AdminLTE/skin-blue settings.

### Before submitting a pull-request to Gitbucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [x] rebased my branch over master
- [x] verified that project is compiling
- [x] verified that tests are passing
- [ ] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [ ] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

PR #1491, opened 2017-03-13
Add Cache-Control header for _avatar and /plugin-assets (refs #777)

kounoike:
For reduce network connection and CPU usage, add Cache-Control header.

On my PC, _avatar image transfer requires ~30ms per user.
/plugin-assets requires 200~300ms per file.
on #1489, some CSS/JS moved to /plugin-assets, it requires maybe 1s.
Because browser caches these files, it makes good effect to CPU usage too.

max-age=3600 means 60*60sec. There is no reason to chose this time.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/AccountController.scala
+++ b/src/main/scala/gitbucket/core/controller/AccountController.scala
@@ -149,6 +149,7 @@ trait AccountControllerBase extends AccountManagementControllerBase {
 
   get("/:userName/_avatar"){
     val userName = params("userName")
+    response.setHeader("Cache-Control", "max-age=3600")
     getAccountByUserName(userName).flatMap(_.image).map { image =>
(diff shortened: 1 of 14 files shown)


takezoe:
This fix looks good, but it might confuse users (or plugin developers may got confused in some cases) when avatar images or plugins are updated because these resources are not available immediately. It may be also necessary a capability to add hash to the url of these resources.

kounoike:
I think add deprecated annotation to old assets(implicit context: Context): String is better for plugin developer.
Should use new assets(path: String)(implicit context: Context): String.

In #1489, I add pluginAssets(). It should be rewrite similary.

takezoe:
> I think add deprecated annotation to old assets(implicit context: Context): String is better for plugin developer.

@kounoike Thanks. I will add several fixes including this after merging.

takezoe:
Fix #1491

takezoe:
Fix #777

PR #1507, opened 2017-03-19
Extending markdown code block renderer by plugin system

nus:
This changes provides code render in a markdown file. A plugin will be able to render a code block like the below code in markdown.

    ```plantuml
    @startuml
    Bob->Alice : hello
    @enduml
    ```

This code will be handle by a plugin.

### DIFF
--- a/src/main/scala/gitbucket/core/plugin/MarkdownCodeRenderer.scala
+++ b/src/main/scala/gitbucket/core/plugin/MarkdownCodeRenderer.scala
@@ -1,0 +1,14 @@
+package gitbucket.core.plugin
+
+import io.github.gitbucket.markedj.Options
+
+/**
+ * A render engine to render code in a makrdown to String.
+ */
+trait MarkdownCodeRenderer {
+
+  /**
+   * Render the given code to String.
+   */
+  def renderCode(code: String, lang: String, escaped: Boolean, options: Options): String
+}
--- a/src/main/scala/gitbucket/core/plugin/MarkdownCodeRenderer.scala
+++ b/src/main/scala/gitbucket/core/plugin/MarkdownCodeRenderer.scala
@@ -1,6 +1,7 @@
 package gitbucket.core.plugin
 
 import io.github.gitbucket.markedj.Options
+import io.github.gitbucket.markedj.Utils._
 
 /**
  * A render engine to render code in a makrdown to String.
@@ -11,4 +12,11 @@ trait MarkdownCodeRenderer {
    * Render the given code to String.
    */
   def renderCode(code: String, lang: String, escaped: Boolean, options: Options): String
+}
+
+object DefaultMarkdownCodeRenderer extends MarkdownCodeRenderer {
+  override def renderCode(code: String, lang: String, escaped: Boolean, options: Options): String = {
+    "<pre class=\"prettyprint" + (if(lang != null) s" ${options.getLangPrefix}${lang}" else "" )+ "\">" +
+    (if(escaped) code else escape(code, true)) + "</pre>"
+  }
 }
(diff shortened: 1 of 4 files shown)


uli-heller:
I like this idea. Could you please explain how a plugin is registered for this?
I assume that "plantuml" somehow related to the plugin, does it?

nus:
@uli-heller
Thanks for your comment.

> I assume that "plantuml" somehow relates to the plugin, doesn't it?

Yes, your assumption is right. That code is an example of this pull request. Instead of "plantuml", kind of code can be available in "scala", "python", or, something, like the below markdown code block.

    ```scala
    object HelloWorld {
      def main(args: Array[String]) {
        println("Hello world.")
      }
    }
    ```



This code can be handled by this change with the below plugin. The plugin code returns highlight of markdown code block or result of evaluating it.

#### Plugin.scala

```scala
import gitbucket.core.plugin.MarkdownCodeRenderer
import io.github.gitbucket.markedj.Options

class Plugin extends gitbucket.core.plugin.Plugin {

  override val markdownCodeRenderers = Seq(
      ("scala", new MarkdownCodeRenderer {
        override def renderCode(code: String, lang: String, escaped: Boolean, options: Options): String = {
          // Returns highlight of the code argument,
          // or a result of evaluating it.
          ...
        }
      })
    )
}

```

nus:
`markdownCodeRenderers` is a class member value of `gitbucket.core.plugin.Plugin`. The member is added in this pull request commit fba2fd461a339d8d879d33ee67802edc28e50b9c .

takezoe:
This looks a useful extension point but I'm cautious about introduce our own syntax to markdown. We have to consider well because it's impossible to remove if it's introduced once. I would like to hear opinion of users about this.

And it have to be possible to be used from renderers other than markdown. So `MarkdownCodeRenderer` is not suitable name.

aadrian:
@takezoe 
> I'm cautious about introduce our own syntax to markdown. 
+1.

Others seem to have the same problem:
https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8588 
https://gitlab.com/gitlab-org/gitlab-ce/issues/4048

http://stackoverflow.com/questions/32203610/how-to-integrate-uml-diagrams-into-gitlab-or-github


> I would like to hear opinion of users about this.

The feature on itself is cool and useful.

The **most important** part however is to be able to generate from the same Markdown files the same HTML without GitBucket too, e.g. with an Gradle/Maven/ANT task.

t-tsutsumi:
@takezoe 
> I'm cautious about introduce our own syntax to markdown.

+1

Although this may be useful, the GitBucket specific dialect will make it difficult to view or edit markdown when using tools other than GitBucket. Also, it will be difficult to upgrade the Markdown (CommonMark) libraries.

In this case, I think that it is better for GitBucket’s plugin to provide a mechanism like [PlantUML Gist](http://uml.mvnsearch.org/).

lefou:
I don't like the idea of a renderer-specific extension point in the Plugin API.

The Plugin API already allows registrations of alternative renderers. E.g. there is an Asciidoctor plugin rendering AsciiDoc. I think, you should first try to provide an alternative Markdown-Plugin providing your additions. Not knowing for sure, but it might be necessary to move the built-in renderer to a (default) plugin first, as a preparation. As the current implementation does not support multiple renderers for the same type.

You could also try to provide some extension capabilities in that plugin itself, if you really think, this must be plug-able.

McFoggy:
> You could also try to provide some extension capabilities in that plugin itself, if you really think, this must be plug-able.

That was the goal of https://github.com/gitbucket/gitbucket/pull/1323 : switch to commonmark which is extensible with plugins ; but I was too ambitious on my capabilities/time to accomplish that.

> but it might be necessary to move the built-in renderer to a (default) plugin first

Would be the cleanest thing to do first.

nus:
Thank you for comments.
But excuse me, I close this MR.
I have no motivation for this MR.

PR #1518, opened 2017-04-01
Add caution message when using H2 database

kounoike:
This PR adds messages when using H2 database. Because H2 is not stable for production environment.

In sign in dialog:
192658-sign in - google chrome

In System settings:
214653-system settings - google chrome

### DIFF
--- a/src/main/twirl/gitbucket/core/admin/system.scala.html
+++ b/src/main/twirl/gitbucket/core/admin/system.scala.html
@@ -1,4 +1,5 @@
 @(info: Option[Any])(implicit context: gitbucket.core.controller.Context)
+@import gitbucket.core.util.DatabaseConfig
 @gitbucket.core.html.main("System settings"){
   @gitbucket.core.admin.html.menu("system"){
     @gitbucket.core.helper.html.information(info)
@@ -20,7 +21,17 @@
             </tr>
             <tr>
               <td>DATABASE_URL</td>
-              <td>@gitbucket.core.util.DatabaseConfig.url</td>
+              @if(DatabaseConfig.url.startsWith("jdbc:h2:")) {
+              <td class="danger">
+                <p>@gitbucket.core.util.DatabaseConfig.url</p>
+                <p>H2 database should be used for evaluation purpose only.</p>
+                <p>
+                  Please configure using external database explained <a href="https://github.com/gitbucket/gitbucket/wiki/External-database-configuration">here</a>.
+                </p>
+              </td>
+              }else{
+                <td>@gitbucket.core.util.DatabaseConfig.url</td>
+              }
             </tr>
           </table>
           <!--====================================================================-->
--- a/src/main/twirl/gitbucket/core/admin/system.scala.html
+++ b/src/main/twirl/gitbucket/core/admin/system.scala.html
@@ -24,7 +24,7 @@
               @if(DatabaseConfig.url.startsWith("jdbc:h2:")) {
               <td class="danger">
                 <p>@gitbucket.core.util.DatabaseConfig.url</p>
-                <p>H2 database should be used for evaluation purpose only.</p>
+                <p>Embedded H2 database should be used for evaluation purpose only.</p>
                 <p>
                   Please configure using external database explained <a href="https://github.com/gitbucket/gitbucket/wiki/External-database-configuration">here</a>.
                 </p>
--- a/src/main/twirl/gitbucket/core/signin.scala.html
+++ b/src/main/twirl/gitbucket/core/signin.scala.html
@@ -1,6 +1,7 @@
 @(userName: Option[Any] = None,
   password: Option[Any] = None,
   error: Option[Any] = None)(implicit context: gitbucket.core.controller.Context)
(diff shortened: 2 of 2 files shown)


aadrian:
> This PR adds messages when using H2 database. Because H2 is not stable for production environment.

Sorry but this is simply not true.
http://h2database.com/html/download.html
We are using H2 since years with hundreds of GB of data without problems.

However when used in embedded mode, several aspects needs to be considered (but this is true for all embedded DBs)

kounoike:
@aadrian Should I change message from "H2" to "embbeded H2"?

- Someone says "GitBucket broken(using H2)" in Japanese twitter.
- [SonarQube](https://www.sonarqube.org/) also shows message "Embedded database should be used for evaluation purpose only".

aadrian:
@kounoike 

>  Should I change message from "H2" to "embbeded H2"?

Yes.  

As it's visible here: http://h2database.com/html/cheatSheet.html  one can use other modes and parameters too.

However there's nothing wrong with embedded mode if used correctly.

Also please note that some the "issues" GitBucket had in the **past** when using H2 were own-made:
  1. DB Model not optimized at all
  2. DB Access Layer - Slick
  3. Production grade connection pool only pretty recently added - HikariCP

Other performance "red flags" we observed with GitBucket installations:
 - GITBUCKET_HOME not on a fastest disk, but in user's home that even gets synced over network
 - no tuning of connection parameters whatsoever (even if the load is high): e.g. bigger connection pool, etc.
  - even if it's there how, no use of LFS (lot's of binaries in repos, because they're old and were converted from Subversion)
  - when hosted under Windows, antivirus not configured not to overload Java apps and files.

> Someone says "GitBucket broken(using H2)" in Japanese twitter.

Searching for "broken" + Any DB will give you results :) .


E.g. we are using even this https://github.com/takezoe/mirage with ~200GB of data without any issues :) .

IMHO there are a few things that can be done in the future for this matter:
 - document what the users need to send us in order to be able to help them with the performance problem (e.g. DB Snapshot, etc.)
 - better default logging for DB operations, so that users can send us useful logs, not just "it's broken" or "it's slow" :) . If it's possible even with a simple click of a button in the admin UI.
 - address **1.** and **2.** from above

rloutrel:
I am always a bit concerned about security.

May I ask to avoid putting information about the infrastructure into the not logged in part?

I am actually thinking, that it should be visible only for logged in users with administration rights.

aadrian:
> I am always a bit concerned about security.

H2 Database can be encrypted, and admins are encouraged to do so: https://github.com/gitbucket/gitbucket/wiki/External-database-configuration#database-encryption

> May I ask to avoid putting information about the infrastructure into the not logged in part?

Users should not be able to access Application Server logs.

What I propose is to have in Admin a button or settings, so that admins if have performance problems or errors, to have an automatic way of sending relevant information to us in order to be able fix them. Otherwise any form of "support" is not really possible and it's just speculation.

kounoike:
I didn't discus performance and security. I concerned about reliability.
Now users can say "GitBucket is shit! Because I lost my source code".
I want to change it to "Data was broken. But it's my problem".

> Searching for "broken" + Any DB will give you results :) .

OK. I googled it.

|DB|all|broken+DB name|%|
|--|---|----|--|
|h2 database|384000|78800|20.5%|
|mysql|151000000|755000|0.5%|
|postgresql|19400000|426000|2.2%|

> E.g. we are using even this https://github.com/takezoe/mirage with ~200GB of data without any issues :) .

Is it embedded H2? Is it reliable with kill -9 <pid>?

aadrian:
> Now users can say "GitBucket is shit! Because I lost my source code".

Sorry but this is total BS. GitBucket is not saving the sources in the DB :) . Even if the entire DB would be lost, **sources**, pages, wiki and attachments would be still there in the ```GITBUCKET_HOME``` directory. 

> Is it embedded H2?

Yes, too. We replicate the data over other nodes, and open there in embedded mode too (e.g. by a desktop application for analytics and offline mode - so without direct DB connection to the server version)

> Is it reliable with kill -9 ?

If it was killed while writing data, that transaction was of course not committed. If the file is corrupt because of the kill than there's a recovery mode (like with most DBs) that must be performed afterwards http://h2database.com/javadoc/org/h2/tools/Recover.html

Hosting GitBucket in Tomcat (correctly configured) or something more proper for production (i.e. **not** Jetty), would greatly reduce the need to ```kill -9``` since it has a working ```shutdown.sh``` script.

kounoike:
> Sorry but this is total BS. GitBucket is not saving the sources in the DB :) . Even if the entire DB would be lost, sources, pages, wiki and attachments would be still there in the GITBUCKET_HOME directory.

Yes, of course I know it. But can you explain to **all** users?

takezoe:
From I and some GitBucket user's experience, when GitBucket crashes by some reason such as OOM, H2 data may be broken. In fact, since it's leading complaints, we should explain them that what they should do to keep their data safety. One solution is giving enough heap to GitBucket and taking H2 backup, or another one is switching to an external database.

I think it's good to show the appropriate explanation and link to the document about external database configuration at the administration console as a first step.

PR #1526, opened 2017-04-05
update dependencies and wording

aadrian:
### Before submitting a pull-request to Gitbucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [x] rebased my branch over master
- [x] verified that project is compiling
- [x] verified that tests are passing
- [ ] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [ ] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

PR #1533, opened 2017-04-09
Dependency updates

aadrian:
Separate #1526 in 2 PRs . This is the second one!


altrive:
It seems `postgresql-embedded` `1.23` have JDK9 compatibility issue.
I've confirmed ci test passed when using version `1.16` 

### Error Message
> 08:01:27.175 [pool-1-thread-1-ScalaTest-running-GitBucketCoreModuleSpec] ERROR r.y.q.e.postgresql.PostgresProcess - Could not use the configured artifact store for cmd, falling back to default InitDb
> java.lang.reflect.InaccessibleObjectException: Unable to make field private int java.lang.reflect.Field.modifiers accessible: module java.base does not "opens java.lang.reflect" to unnamed module @351087a3
> at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:207)
> at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:171)
> at java.base/java.lang.reflect.Field.setAccessible(Field.java:165)
> at ru.yandex.qatools.embed.postgresql.util.ReflectUtil.setFinalField(ReflectUtil.java:24)
> ...

### Root cause of the problem
It seems this JDK9 compatibility issue is caused by following problem.
<http://stackoverflow.com/questions/41265266/how-to-solve-inaccessibleobjectexception-unable-to-make-member-accessible-m>

project/plugins.sbt
```diff
@@ -6,3 +6,4 @@ addSbtPlugin("com.earldouglas"      % "xsbt-web-plugin"   % "2.1.1")
 addSbtPlugin("fi.gekkio.sbtplugins" % "sbt-jrebel-plugin" % "0.10.0")
 addSbtPlugin("com.typesafe.sbt"     % "sbt-pgp"           % "0.8.3")
 addSbtPlugin("io.get-coursier"      % "sbt-coursier"      % "1.0.0-M15")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
```
takezoe: I prefer that helper plugin like this is defined in user local sbt settings (`~/.sbt/0.13/plugins/build.sbt `).

aadrian: > I prefer that helper plugin like this is defined in user local sbt settings

Done!

build.sbt
```diff
@@ -54,9 +54,9 @@ libraryDependencies ++= Seq(
   "javax.servlet"                   %  "javax.servlet-api"            % "3.1.0"          % "provided",
   "junit"                           %  "junit"                        % "4.12"           % "test",
   "org.scalatra"                    %% "scalatra-scalatest"           % ScalatraVersion  % "test",
-  "org.mockito"                     %  "mockito-core"                 % "2.7.16"         % "test",
+  "org.mockito"                     %  "mockito-core"                 % "2.7.22"         % "test",
   "com.wix"                         %  "wix-embedded-mysql"           % "2.1.4"          % "test",
-  "ru.yandex.qatools.embed"         %  "postgresql-embedded"          % "1.14"           % "test"
+  "ru.yandex.qatools.embed"         %  "postgresql-embedded"          % "1.23"           % "test"
```
takezoe: Does it work with Java9 if postgresql-embedded is rolled back? How about rolling back it once?

aadrian: > Does it work with Java9 if postgresql-embedded is rolled back?

It seems that ```postgresql-embedded``` does not work, nor is supposed to work with Java 9.

It's not even tested with Java 9:
https://github.com/yandex-qatools/postgresql-embedded/blob/master/.travis.yml

I tried to run it with Java 9 but can't get it work so far:
https://travis-ci.org/aadrian/postgresql-embedded/jobs/227742380

The fact that some older version in combination with other older libraries worked with Java 9 is just a pure coincidence IMHO.

I raised an issue https://github.com/yandex-qatools/postgresql-embedded/issues/76 but so far no answer/solution.

Not sure why GitBucket must work with Java 9, when most libraries it depends on, don't even test against it, e.g. it's own libraries like https://github.com/gitbucket/solidbase/blob/master/.travis.yml  :) .

takezoe: > I raised an issue yandex-qatools/postgresql-embedded#76 but so far no answer/solution.

I've found that issue. Good job 👍 

> Not sure why GitBucket must work with Java 9

Some users are already trying to use GitBucket with Java 9. I'm not daily testing GitBucket with Java 9, but I would like to run test cases on Travis at least.

aadrian: Java 9 should work now (at least now). Please merge.

> Some users are already trying to use GitBucket with Java 9.

I understand that, but IMHO this does not belong in the master branch (maybe a Java 9 branch until Travis and all other support Java 9 officially too).

build.sbt
```diff
@@ -25,26 +25,26 @@ libraryDependencies ++= Seq(
...
-  "com.novell.ldap"                 %  "jldap"                        % "2009-10-07",
-  "com.h2database"                  %  "h2"                           % "1.4.192",
-  "mysql"                           %  "mysql-connector-java"         % "5.1.39",
-  "org.postgresql"                  %  "postgresql"                   % "9.4.1208",
-  "ch.qos.logback"                  %  "logback-classic"              % "1.1.7",
-  "com.zaxxer"                      %  "HikariCP"                     % "2.4.6",
-  "com.typesafe"                    %  "config"                       % "1.3.0",
-  "com.typesafe.akka"               %% "akka-actor"                   % "2.4.12",
+  "joda-time"                       %  "joda-time"                    % "2.9.9",
+  "com.novell.ldap"                 %  "jldap"                        % "4.3",
```
xuwei-k: 4.3 is older version?

- http://repo1.maven.org/maven2/com/novell/ldap/jldap/
- http://repo1.maven.org/maven2/com/novell/ldap/jldap/maven-metadata.xml

screen shot 2017-05-03 at 14 13 33

aadrian: > 4.3 is older version?

According to VersionEye :) (what I used)
https://www.versioneye.com/user/projects/590a19855420bc001227f58b

The numbering doesn't seems OK however, since from the sources the most recent version should be 4.6  http://www.openldap.org/devel/gitweb.cgi?p=openldap-jldap.git;a=commit;h=bae9f556e97ad5bb0f76fe897c9b298a6021e244

diogosantana: @takezoe Why GitBucket use novell ldap at all? Why not use plain old JNDI?

I was checking the LDAPUtil.scala file and at some point it changes the trustStore for the whole JVM:
`System.setProperty("javax.net.ssl.trustStore", keystore)`

PR #1517, opened 2017-04-01
Fix #1516 send noimage.png if user is removed.

kounoike:
This PR fixes #1516.
Sorry, I didn't care to removed user in #1497.

Before:
205518-manage users - google chrome

After:
205802-manage users - google chrome

### Before submitting a pull-request to Gitbucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [ ] rebased my branch over master
- [x] verified that project is compiling
- [x] verified that tests are passing (mysql test is failed by timeout)
- [ ] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [x] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

PR #1538, opened 2017-04-13
Fix loading avatar from resources

dbronecki:
Unknown users avatars (the ones who don't have account created in Gitbucket) wasn't displaying (404 error).

After investigating, I found that resource path was invalid.

Before:
before

After:
after


dbronecki:
I'm sorry, it seems to be working in master without my changes. I was comparing with recent stable version. It's fixed in #1517

PR #1564, opened 2017-04-29
Disabling Server header on Jetty

t-tsutsumi:
There is no point in returning the `Server` header.

For reference, Tomcat 9.0.x and 8.5.x no longer set the Server header by default.

### DIFF
--- a/src/main/java/JettyLauncher.java
+++ b/src/main/java/JettyLauncher.java
@@ -1,3 +1,5 @@
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.webapp.WebAppContext;
 
@@ -5,6 +7,8 @@ import java.io.File;
 import java.net.URL;
 import java.net.InetSocketAddress;
 import java.security.ProtectionDomain;
+import java.util.Arrays;
+import java.util.Collection;
 
 public class JettyLauncher {
     public static void main(String[] args) throws Exception {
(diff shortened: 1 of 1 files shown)


takezoe:
Conflicting with #1563.

t-tsutsumi:
@takezoe Conflict was resolved.

src/main/java/JettyLauncher.java
```diff
@@ -64,6 +68,11 @@ public static void main(String[] args) throws Exception {
 //        connector.setPort(port);
 //        server.addConnector(connector);
 
+        // Disabling Server header
+        Arrays.stream(server.getConnectors()).map(Connector::getConnectionFactories).flatMap(Collection::stream)
+                .filter(HttpConnectionFactory.class::isInstance).map(HttpConnectionFactory.class::cast)
+                .map(HttpConnectionFactory::getHttpConfiguration).forEach(config -> config.setSendServerVersion(false));
```
takezoe: I don't think that stream API is appropriate in this case. for loop like following is natural and much enough.
```java
for(Connector connector: server.getConnectors()){
    for(ConnectionFactory factory: connector.getConnectionFactories()){
        if(factory instanceof HttpConnectionFactory){
            ((HttpConnectionFactory) factory).getHttpConfiguration().setSendServerVersion(false);
        }
    }
}
```

t-tsutsumi: @takezoe It makes sense. I revised according to your instruction.

takezoe:
@t-tsutsumi Thanks for quick fix!

Issue #1157, opened 2016-03-31
Support displaying user's real name instead of user ID when using Active Directory


sseiichi:
When I using Active Directory, I want to display user's real name instead of user's ID because it's a meaningless employee id number in our organization.

aadrian:
see this: #652 too. It is required basically to allow a greater number of aliases, not just one property and also not when using LDAP.

PaulAdams-Conti:
I would really like to see this being implemented as well but doesn't seem to be planned for any milestone. Not so useful seeing the user's ID. Even when there is an image for the user, it is not so easy to make out who it is sometimes.

Issue #652, opened 2015-03-13
User aliases support!


hansgru:
Please add user aliases support so that several email addresses (and user names) that are present in git commits to be associated and visualized with the same user.

This is required, since over the time, a user will have more than one or or will push from more systems.

This should be editable only by the admin, to prevent misuse.

aadrian:
+1

kt81:
+1

cundd:
Thank you! 🎉

Issue #336, opened 2014-04-02
Always use "display name" in the front end instead "username" for better ldap integration


gianluca-sabena:
I suggest to always use the filed "full name" instead "username" in the interface, this has some advantages:
- user can customize it
- better coherence in the front end, a user is always called with the same name
- it offers a better integration with ldap (username is connected with ldap uid filed and in some company this filed is a serial number, news feed become useless...)

As of today, the news feed use the filed "username", but wiki recent changes use "full name"

PaulAdams-Conti:
Any likelihood of this being implemented. We only have a user id as username and some people don't like having their picture uploaded so it makes it very difficult to assign an issue when faced with a list of uids.

PR #1573, opened 2017-05-03
Add support use mail address as user name when using LDAP

t-tsutsumi:
This PR provides an option to use the part of the mail address that appears to the left of the @ symbol (hereinafter called "PREFIX") as the user name when using LDAP.
If using this option, the user name attribute is used only to identify the user on login.

IMHO, a large bandwidth is required to support user aliases related to #336, #652, and #1157, but in many use-cases, the PREFIX of the mail address registered in LDAP becomes unique and can be used instead of aliases. So this PR could be a good start.

#### 1. System settings
001_system-settings

#### 2. Login
uid=`12345678`, mail=`john-doe@users.noreply.github.com`, sn=`John Doe`
002_login

#### 3. Profile
003_profile

#### 4. News feed
004_news-feed


t-tsutsumi:
I misunderstood why a user can't bind due to misconfiguration of LDAP server. Since this feature can be used now, I will close PR.

PR #1577, opened 2017-05-03
Make read LDAP bind password from environment variables

### LINKED ISSUE #634: LDAP BindDN password is plaintext readable in gitbucket.conf
This is an issue, as in my case, the gitbucket server can be launched by different people, with different LDAP Bind DN.
Here are some solution proposals :
1 :  Store the whole LDAP authentication attributes of the gitbucket.conf into the database, and maybe the SMTP part too.
2 : Allow to pass the BindDN password as command line argument (just like --data from the command line when using "java -jar gitbucket.war --data=/home/gitbucket)  and the BindDN too. 
3 : Crypt the whole gitbucket.conf, with the key stored in database
4 : I could avoid the issue by using an LDAP dedicated accound, for which password being readable would not be problematic  (not sure it is feasible).

Do you have other ideas ?
My favorite would be 2 (command line argument), with the option of passing the Bind DN too.

t-tsutsumi:
I think that a realistic way to protect the LDAP bind password that meets the following minimum requirements is to use environment variables.

1. GitBucket should be able to use with operating system's process manager such as Upstart, systemd, and Task Scheduler. In other words, do not use the interactive interface.
1. System users, processes, and applications that can read the LDAP bind password should be minimal.

It is important to note that encryption does not provide any value unless you use an external security module (e.g. HSM) or agents together.

#### Example with systemd
```bash
cat << '_EOF_' > /etc/systemd/system/gitbucket.service
[Unit]
Description=GitBucket
After=network.target

[Service]
User=gitbucket
Environment="GITBUCKET_LDAP_BIND_PASSWORD=secret"
ExecStart=/usr/bin/java -Xms1024m -Xmx1024m -Djava.net.preferIPv4Stack=true -jar /usr/local/lib/gitbucket/gitbucket.war --gitbucket.home=/var/lib/gitbucket
Restart=always
PrivateTmp=true

[Install]
WantedBy=multi-user.target
_EOF_

chmod 400 /etc/systemd/system/gitbucket.service
systemctl daemon-reload
systemctl enable gitbucket.service
systemctl start gitbucket.service
```

Closes #634.

### DIFF
--- a/src/main/scala/gitbucket/core/util/LDAPUtil.scala
+++ b/src/main/scala/gitbucket/core/util/LDAPUtil.scala
@@ -18,6 +18,7 @@ object LDAPUtil {
   private val logger = LoggerFactory.getLogger(getClass().getName())
 
   private val LDAP_DUMMY_MAL = "@ldap-devnull"
+  private val ENV_NAME_LDAP_BIND_PASSWORD = "GITBUCKET_LDAP_BIND_PASSWORD"
 
   /**
    * Returns true if mail address ends with "@ldap-devnull"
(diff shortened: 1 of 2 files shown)


takezoe:
From other view, it might be nice for docker use as well if all properties of `database.conf` and `gitbucket.conf` are configurable by environment variables (or system properties).

This is just idea, but how do you think?

t-tsutsumi:
@takezoe I think that's a good idea. This makes it possible to create a configurable GitBucket's docker image without a solution like envsubst.

takezoe:
@t-tsutsumi Thanks. I will try to make it in an another branch. It might cover functinallity of this request in other way.

t-tsutsumi:
Solved #1617.

PR #1579, opened 2017-05-04
dropdown helper function improvement & modification

tkgdsg:
Helper can specify placeholder for it's filter.
To identify each dropdown input filter by id, use unique Id.
----
First, I tried following code to initialize input filter id.

`@filterId = @{filter.toString.replaceAll("[^a-zA-Z0-9_]","_") + scala.util.Random.alphanumeric.take(4).mkString }
`

But this code produces different id where filterId is used.
So I abandoned to use this technique.

**If you have any good solution for this, I welcome your comment.**

----

### DIFF
--- a/src/main/twirl/gitbucket/core/helper/dropdown.scala.html
+++ b/src/main/twirl/gitbucket/core/helper/dropdown.scala.html
@@ -2,7 +2,9 @@
   prefix: String  = "",
   style : String  = "",
   right : Boolean = false,
-  filter: String = "")(body: Html)
+  filter: String = "",
+  uniqId: String = scala.util.Random.alphanumeric.take(4).mkString)(body: Html)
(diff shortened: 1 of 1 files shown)


takezoe:
@tkgdsg What's a purpose of this pull request? My original intention is that the caller should pass unique value as `filter`, but do you mean that this fix makes to be unnecessary to think in the caller?

tkgdsg:
This patch enables to specify place holder to each dropdown.
With this patch, #1555 is replaced to use this helper method's native function.

I think developer need not to care about the id of input box, if unique id is assigned automatically in this implementation.
This idea is to re-use filter variable for place holder string.

takezoe:
Thanks for explanation. So `uniqId` may be not necessary as a template argument. I will fix after merge this request.

PR #1581, opened 2017-05-05
Add convenient installation script for Linux systemd

t-tsutsumi:
It provides a means to easily install or upgrade GitBucket on Linux systemd.

I will post it to the wiki after this PR is merged.

### DIFF
--- a/contrib/linux/systemd/README.md
+++ b/contrib/linux/systemd/README.md
@@ -1,0 +1,44 @@
+# Install & Upgrade GitBucket on Linux systemd
+
+It provides a script for easy installation and upgrade of GitBucket using systemd.
+
+## Requirements
+
+This script requires at least:
+* A recent Linux distribution with systemd. The following OS has been confirmed.
+    * Red Hat Enterprise Linux 7.3
+    * Ubuntu 16.04.2 LTS
+* A Java Runtime Environment 8 or later.
+* A sudoer.
+
+## Install (and run)
+
+1. Download the script.  
+`$ curl -O https://raw.githubusercontent.com/gitbucket/gitbucket/master/contrib/linux/systemd/gitbucket-install.sh`
+1. Edit the configurations of the script as you like.
+1. Install the latest GitBucket and run.  
+`$ bash gitbucket-install.sh`
+
+Note: If you want to use an external database, stop the service after the installation is completed and then set it. See also https://github.com/gitbucket/gitbucket/wiki/External-database-configuration.
+
+## Upgrade
+
+It can be executed from the state where the service is running.
+
+1. Upgrade GitBucket to the latest version using the configured script.  
+`$ bash gitbucket-install.sh`
+
+Note: A plugins will not be upgraded, please upgrade manually.
+
+## Tips
+
+1. Start the service  
+`$ sudo systemctl start gitbucket.service`
+1. Stop the service  
+`$ sudo systemctl stop gitbucket.service`
+1. View the logs  
+`$ sudo journalctl -u gitbucket.service `
+1. Uninstall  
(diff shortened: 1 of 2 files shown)


t-tsutsumi:
Since diff in README.md is hard to see, please refer to the following.
https://github.com/t-tsutsumi/gitbucket/tree/pr/add-installation-script-for-linux-systemd/contrib/linux/systemd

aadrian:
@t-tsutsumi , @takezoe  Isn't there a generic and *simpler* way to generate installers/updaters for **all** distributions with SBT? (and also Windows ?)

e.g. For Gradle (with built in Scala support) there's something like:  https://blog.heckel.xyz/2014/05/29/gradle-create-windows-installers-debian-packages-manage-a-ppa-and-optional-sub-projects/

*(or many other alternatives.) *

I've found for **SBT** this: https://github.com/sbt/sbt-native-packager ( with quite good documentation http://sbt-native-packager.readthedocs.io/en/stable/ )  and the "Java Server Application Archetype" seems to be exactly what GitBucket would need besides a WAR distribution.

Than you.

t-tsutsumi:
IMHO, I think that it is unnecessary to use a packager with abundant functions for the following reasons.
* Since GitBucket's distribution has only one WAR file, installation and upgrade are easy.
* When using packages like RPM or DEB, in principle only one can be installed on a system.
* Since there is no bandwidth to provide and maintain YUM and APT repositories in the community, OS native packages have less merit.

@aadrian I do not know much about SBT Native Packager, but is there a function to upgrade from GitHub release like this PR script?

aadrian:
> I think that it is unnecessary to use a packager with abundant functions

Not for the functionality itself that it provides (since it's only a WAR file), **of course** :) . 

However, native installers/updaters/starters would be very good for something else: **exposure** and **advertisement** :) : if users can find GitBucket in all repos/distros/platforms easily, than much more users would be ready to try it out :) .

Also please note that quite a few companies have strange internal policies that allow to install only from repos with "official" distributions - e.g. only RPMs, or only DEB software, or for Windows only MSI that are automatically installed from internal mirrors (so no users that can run directly a ```java -jar gitbucket.war``` ).

>  but is there a function to upgrade from GitHub release like this PR script?

Most native installers provide this functionality, but it's in the tool itself, e.g. ```brew upgrade gitbucket```, so by having those native installers, ```install``` ```run as service``` ```stop``` ```update/upgrade``` and ```uninstall``` would be provided.

t-tsutsumi:
> if users can find GitBucket in all repos/distros/platforms easily, than much more users would be ready to try it out :) .

I agree, I think it would be wonderful if this could be done. However, anyway, we need a solution to realize without effort. 😔

takezoe:
We can create OS packages using sbt-native-packager but it would make additional costs continuously to maintain. We don't have enough resources to do it, and I think that `java -jar gitbucket.war` is easy enough to try.

sjas:
Maybe just run in in a tomcat, as warfiles are intented to?
From an admin's perspective you don't want to run jar's from the commandline, as this will just cause configuration entropy over time, plus dev's need to support that installation part, too.

Just wrote myself a howto for such a setup and just pasted it here, too:
https://github.com/gitbucket/gitbucket/wiki/Installation-on-ubuntu-16.04-with-oracle_java8---tomcat8---mysql---proper-apache_proxy

Didn't put it into the single categories, thought splitting it up might confuse more than help.

aadrian:
@takezoe 
> We can create OS packages using sbt-native-packager but it would make additional costs continuously to maintain. We don't have enough resources to do it

There's not much to maintain, as visible from here https://github.com/sbt/sbt-native-packager
and the reachability of GitBucket would be greatly increased (it even generates automatic Docker stuff - managed manually at the moment for GitBucket).

While I also like the quick `java -jar gitbucket.war` when trying, for production systems is totally unsuitable.

SIkebe:
Now original author of this PR is ***ghost*** account, we should close this?

PR #1584, opened 2017-05-07
dropdown input filter element id from placeholder fix

tkgdsg:
if html.dropdown specified with filter placeholder that includes inhibit character, dropdown filter does not work.
Current, dropdown filter element id is created by placeholder string + random 4 characters.
If placeholder includes inhibit character for html id, (e.g. '\s' , '/' and so on)
input filter element does not recognized by that id.
This patch replace characters(except [a-zA-Z0-9]) to "_"( allowed to use in id ).

### DIFF
--- a/src/main/twirl/gitbucket/core/helper/dropdown.scala.html
+++ b/src/main/twirl/gitbucket/core/helper/dropdown.scala.html
@@ -3,7 +3,7 @@
   style : String  = "",
   right : Boolean = false,
   filter: String = "")(body: Html)
-@defining(if(filter.isEmpty) "" else filter + "-" + scala.util.Random.alphanumeric.take(4).mkString){ filterId =>
+@defining(if(filter.isEmpty) "" else filter.replaceAll("[^a-zA-Z0-9]","_") + "-" + scala.util.Random.alphanumeric.take(4).mkString){ filterId =>
   <div class="btn-group" @if(style.nonEmpty){style="@style"}>
     <button
         class="dropdown-toggle btn btn-default btn-sm" data-toggle="dropdown">


takezoe:
This parameter is used for id attribute basically, so requiring suitable value for that purpose to caller is reasonable. In my opinion, checking in the template side is overkill.

takezoe:
`@filter` which is used as placeholder may be replaced with `@value`:
```html
<li><input id="@filterId-input" type="text" class="form-control input-sm dropdown-filter-input" placeholder="@filter"/></li>
```
I didn't intended that  `filter` is used as placeholder originally.

tkgdsg:
OK.

I propose following alternatives.
Give me a comment please.

1. just revert PR #1579:  placeholder string is fixed as "Filter"
2. create new variable for placeholder
```
@(value : String  = "",
  prefix: String  = "",
  style : String  = "",
  right : Boolean = false,
  filter: String = "",
  placeholder: String = "Filter")(body: Html)
```
  if filter.nonEmpty, filter element is created and default placeholder is "Filter".
3. create a tuple for filter id and placeholder
```
@(value : String  = "",
  prefix: String  = "",
  style : String  = "",
  right : Boolean = false,
  filter: (String, String) = ("", "Filter"))(body: Html)
```
  call like this
```
@gitbucket.core.helper.html.dropdown(
    originRepository.owner + "/" + originRepository.name, 
    "base fork","",false,("baseRepo","Find repository ..."))
```
  And off-course all this "dropdown" function caller is modified to this expression.

takezoe:
@tkgdsg Sorry for my slow reply.

> 3. create a tuple for filter id and placeholder

This looks good to me. The initial value of placeholder string isn't necessary because it can't be omit when calling.

tkgdsg:
Thank you for your response.

I will create a new pull request, based on the idea option 3.

Issue #1430, opened 2017-01-20
use webjars


nazoking:
Would you like to use webjars?
http://www.webjars.org/

Webjars has libraries of npm and bower registered, you can easily use it from servlet by just registering it in sbt. Many registered javascript and css are optimized. (refs #965 Compress JavaScript and CSS resources)

Also, by using webjars, it is guaranteed that the original is unchanged, so it helps to upgrade the library version.

aadrian:
using npm and webpack directly (from gradle or sbt) offer all the advantages and many more IMHO.

McFoggy:
Is this issue to get rid of `gitbucket/src/main/webapp/assets/vendors/` and to allow somehow an automatic dependency managment towards JS/CSS libs during buildtime?

aadrian:
> Is this issue to get rid of gitbucket/src/main/webapp/assets/vendors/ and to allow somehow an automatic dependency managment towards JS/CSS libs during buildtime?

If this is the case, it can be achieved wihout WebJars very easily: just have a simple ```package.json``` with the needed dependencies, and let NPN take care of the download und update. Than with WebPack https://webpack.js.org/ the usage can  be optimized (much better than it can be with WebJars) .

nazoking:
> using npm and webpack directly (from gradle or sbt) offer all the advantages and many more IMHO.

I agree. But webjars are very simple. There is no need to set anything. Npm and webpack require new software and configuration files for development. It is somewhat complicated to use from sbt, exhausting developers. If we aim for Single Page Application, webpack looks best. However, webjars works well and it is easy to migrate if we use the current usage of javascript.

nazoking:
> Is this issue to get rid of gitbucket/src/main/webapp/assets/vendors/ and to allow somehow an automatic dependency managment towards JS/CSS libs during buildtime?

yes. i would rid gitbucket/src/main/webapp/assets/vendors/ .

nazoking:
webjars is not software. it is library for scala or java, servlet. We need no sbt-plugins.

The work is all just the following.

  1. add script name to `dependencies` in `build.sbt`
       * like `"org.webjars.nmp" % "bootstrap" % "3.3.6"`
  2. grep and rewrite path in script or link tag.
[...]

PR #1595, opened 2017-05-13
[WIP] Experimental use WebJars

t-tsutsumi:
This is a prototype for using WebJars with GitBucket. This feature was proposed in #1430.

Currently, I know that there are the following problems.

* The diff feature does not work. (Not investigated)
* WAR becomes bulky.
    * Since WebJars includes unnecessary resources such as documents, the size of the WAR has been greatly expanded. It would be desirable to remove unnecessary files and repackage them.

Anyone who has bandwidth to work on above problem is very welcomed.


aadrian:
> It would be desirable to remove unnecessary files and repackage them.

Sorry but this is the biggest disadvantage of WebJars - they work only for very small scenarios.
IMHO much better is to use WebPack or some other NodeJS based solution for the JS/Css resources.

t-tsutsumi:
> IMHO much better is to use WebPack or some other NodeJS based solution for the JS/Css resources.

I agree. Would be great to see an initial prototype of using WebPack to see a scale of this change.

aadrian:
> Would be great to see an initial prototype of using WebPack to see a scale of this change.

This GitBucket plug-in https://github.com/tomoki1207/gitbucket-explorer-plugin is making use of this concept directly.

Basically there's only the NPM project file: https://github.com/tomoki1207/gitbucket-explorer-plugin/blob/master/package.json  that has everything.

ritschwumm:
i don't get it. why tie yourself to a whole load of complex and error-prone technology when you can get what you need in a much simpler way?

aadrian:
> when you can get what you need in a much simpler way?

There's no simpler way :( . JS and CSS dependencies are handled correctly (like Maven or SBT does) only by NPM, and tools build upon it only can refine out of a library the required/referenced parts.
Also only that that way transitive dependencies and conflicts are handled correctly. 

As I mentioned in https://github.com/gitbucket/gitbucket/issues/1430 once configured, the user/developer does not need to see any part of it, since there are SBT or Gradle plug-ins that hide everything, and the output is the correct but minimal set of files that contain only the JS and CSS code that is directly referenced.

nazoking:
jsdiff was changed from original.
One of the objectives of my proposal #1430  is to make it possible to distinguish between those that have been changed and those that are not.

t-tsutsumi:
> jsdiff was changed from original.

I tried rolling back the jsdiff before creating this PR, but the diff feature did not work. Probably there are other problems as well.

SIkebe:
Now original author of this PR is ***ghost*** account, we should close this?
I think using webpack (#1997) is better, too.

aadrian:
> Now original author of this PR is ghost account, we should close this?

@SIkebe Yes, IMHO we should close this PR.

PR #1487, opened 2017-03-11
Plugin hot deployment

takezoe:
See #1484 and #1468

PR #1627, opened 2017-07-02
[WIP] Copy bundled plugins in initialization process

### LINKED ISSUE #1625: Provides a way to bundle plugins to GitBucket distribution

takezoe:
The purpose of this pull request is making posisble to bundle default plugins with the GitBucket distribution.

The basic concept is copying bundled plugin jars to `GITBUCKET_HOME/plugins` before plugin initialization, but there are two concerns to be considered:

- The current implementation needs to include plugin jars in the gitbucket jar which published to the Maven Central repository. It's desireble to be included in only gitbucket.war.
- Bundled plugins are always enabled. It's impossible to disable. It might have to make the plugin hot deployment (#1487) to disable bundled plugins.

Close #1625


takezoe:
Merged with #1487

PR #1666, opened 2017-08-08
Fix lint and typos

HairyFotr:


takezoe:
@HairyFotr Thanks a lot!

PR #1672, opened 2017-08-14
Add support to MIME Text part in Email notification.

kounoike:
I use [Becky! 2](http://www.rimarts.co.jp/becky.htm) for receiving E-mail at work. My company restricts that don't use rendering HTML mail in Becky. 

So, if commented below:
```
#1 #2 #3 #4 #5

**abc** ~~def~~ `ghi`

` ` `python
import sys
print 42
print "abc"
 ` ` `

- this
-  is
    - list


1. ordered
1. list
```

it shows in Becky:
20170814-145216 - nas - becky

It is hard to understand comment.

Then I want to add original markdown as MIME Text part in gitbucket-notification-plugin. First step, I want to add MIME Text part support in GitBucket's Notifier.

In this PR, I added `textMsg: Option[String]` to `toNotify` and `send`. and call it with `None` by previous signature's one. It doesn't break plugin interface.


takezoe:
@shimamoto Could you review this pull request?

src/main/scala/gitbucket/core/util/Notifier.scala
```diff
@@ -19,8 +19,12 @@ import SystemSettingsService.Smtp
  * Please see the plugin for details.
  */
 trait Notifier {
+  def toNotify(subject: String, htmlMsg: String)
+    (recipients: Account => Session => Seq[String])(implicit context: Context): Unit = {
+    toNotify(subject, htmlMsg, None)(recipients)
+  }
 
-  def toNotify(subject: String, msg: String)
+  def toNotify(subject: String, htmlMsg: String, textMsg: Option[String])
```
shimamoto: @kounoike thanks

> It doesn't break plugin interface.

As you say, it is supposed to be this way as it should be, but... we might as well add `textMsg: String` just this once if you fix the plugin right after this. How you do you feel about it?

kounoike: It can be used by other plugin.

shimamoto: Well...either way, this fix affects other plugins...

@takezoe Well, you decide.

takezoe: Actually, `htmlMsg` might have to be optional, not `textMsg`. We can correct APIs in this timing because there aren't so many plugins for now.

kounoike:
I'm sorry. I have operated erroneously.
I rewrite `toNotify` interface as:

1. `textMsg: String` is required
2. `htmlMsg: Option[String]` is optional.

and `SystemSettingsController` use `Mailer.send` directory, it fixed too.

takezoe:
LGTM!

PR #1675, opened 2017-08-14
Handle errors to show errors prettify and logging it.

kounoike:
Now, GitBucket doesn't handle exceptions. It handled by scalatra default handler. When error thrown, error description and stack trace only displayed to user. There is no log. In some case, administrator can't see it. Then, administrator can't solve problem.

In this PR, handle Exceptions by `error {}` to show errors prettify and log it by logger.


takezoe:
It looks nice feature because it would help debug and support. 😉 

However it's not preffered that displaying starcktrace on browser in default from the security view.
GitBucket should provide an option to switch it at least.

kounoike:
OK, I will add such feature. and I propose one idea:
For administrator, display stack trace always even if option switch is disabled.

kounoike:
@takezoe  OK, I commit.

for non-administrator or unauthenticated:
20170817-174631 - error - google chrome

for administrator or debug is true:
20170817-174637 - error - google chrome

each case, write error to logger:

takezoe:
@kounoike This pull request seems to conflict with #1676 which is merged before.

kounoike:
@takezoe OK, merged it!

takezoe:
@kounoike Thanks for great contribution! This will decrease our support cost a lot :-)

takezoe:
@kounoike I can't find how to configure the debug setting in the settings page. Did you forget to commit some parts of this fix?

kounoike:
Sorry, I forgot to explain. I didn't write UI settings because it should be temporally settings by system property or environment settings. It introduced at #1617.

PR #1697, opened 2017-09-06
Capitalize GitBucket in plugin description

ryoon:
Could you fix a comment in Pages plugin description?

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -41,7 +41,7 @@
   {
     "id": "pages",
     "name": "Pages Plugin",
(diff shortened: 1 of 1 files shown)


takezoe:
This came from the plug-in definition in gitbucket-pages-plugin:
https://github.com/gitbucket/gitbucket-pages-plugin/blob/master/src/main/scala/Plugin.scala#L12

ryoon:
I will submit a patch for https://github.com/gitbucket/gitbucket-pages-plugin .
Thank you.

Issue #1702, opened 2017-09-11
Transfer to URL with commit ID by "y" key


takezoe:
GitHub (and also GitLab) supports keyboard shortcut `y` which transfers a browser to the URL with commit ID. GitBucket should support it.

xuwei-k:
FYI, there are some keyboard shortcuts other than "y"
https://help.github.com/articles/using-keyboard-shortcuts/

takezoe:
Thanks. Only `t` has been supported currentry.

PR #1703, opened 2017-09-12
Change URL with commit ID when user selects line(s).

kounoike:
related to #1702.
In GitHub, if user select lines in blob view, user needs to switch commit URL by 'y' key. Because it maybe changes after some commits pushed to the branch.
But, I assume '/user/repo/blob/branch/path/to/file.ext#L1-L3' link is BAD URL. It should be changed to URL with commit ID automatically.
This PR implements such behavior.


takezoe:
@kounoike Nice hack! 👍

PR #1726, opened 2017-09-29
Add additional system paths to reserved names

reap3r119:


takezoe:
LGTM 👍

aadrian:
@takezoe can't Scalatra fetch this backwards? I.e. deliver a list with all servlet mappings, so that they won't be associated with orgs? Considering that users also have (different) plug-ins, "automatic" solution would also blacklist those.

PR #1733, opened 2017-10-09
Removes imports of external links so that gitbucket can be used in off-internet-environments

masterwto:
### Before submitting a pull-request to GitBucket I have first:

- [] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [] rebased my branch over master
- [] verified that project is compiling
- [] verified that tests are passing
- [] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

PR #1736, opened 2017-10-11
Add local version of Source Sans Pro

reap3r119:
Include the Source Sans Pro font family locally instead of importing it from Google Fonts. This is an alternative to removing the font (#1733)


masterwto:
I think the sources should be re-organised into `/vendor/google-fonts/sans-pro` folder instead of `/common/fonts/`.

PR #1757, opened 2017-10-25
Rename PluginRegistory.scala to PluginRegistry.scala

reap3r119:


takezoe:
Hehe, my typo. Thanks for fix 👍

PR #1786, opened 2017-12-01
fix redirect URL path encode

guyon:

- [x]  read the [contribution guidelines](htt ps://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [ ]  rebased my branch over master
- [x]  verified that project is compiling
- [x]  verified that tests are passing
- [ ]  squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [ ]  [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

### DIFF
--- a/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
+++ b/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
@@ -321,7 +321,7 @@ trait RepositoryViewerControllerBase extends ControllerBase {
       commit      = form.commit
     )
 
-    redirect(s"/${repository.owner}/${repository.name}/blob/${form.branch}/${
+    redirect(s"/${repository.owner}/${repository.name}/blob/${urlEncode(form.branch)}/${
       if (form.path.length == 0) urlEncode(form.newFileName) else s"${form.path}/${urlEncode(form.newFileName)}"
     }")
   })


ritschwumm:
what about repository-owner and -name? are they safe?

guyon:
@ritschwumm 

Repository name and owner name is valid check input filed  when create.
But, to encode more  safety and better.

validator

takezoe:
> But, to encode more safety and better.

Extacly, but it affects entire of codebase. I'm not sure whether we should do it.

Anyway, this pull request looks good! I will merge after 4.19.0 release because 4.19.0 codebase has been freeezed already.

ritschwumm:
even if names are filtered on input, it's never a good idea to rely on it.
when you put something in an URL (or HTML), escape it.

takezoe:
@ritschwumm Of course, they are escaped in HTML. I talked about only repository owners and names which come from the database and used for redirection.

I agree with you that they should be escaped, but it's safe basically and we don't have enough time and we have a lot of tasks to do other than it. So I can't give so high priority to do it, but creating another issue or pull request (better than issue ;-)) is welcome!

ritschwumm:
well.. personally, i tend to fix this kind of stuff as early a possible -
people tend to copy this kind of code and multiply the problem over time.

PR #1802, opened 2017-12-10
Add a pulldown menu to choose the merge strategy

### LINKED ISSUE #898: allow to merge PR from UI "fast forward"
Would be great to allow users to choose when accepting a PR if they want to merge with:
- `--no-ff` (which is the default behavior today of gitbucket & github)
- `--ff-only` fast forward only
- `--ff` making fast forward when possible and merge commit otherwise

as see git [merge documentation](http://git-scm.com/docs/git-merge#_options).

PS: We could even think of another possibility which is a rebase.

takezoe:
merge-strategy

Closes #898


McFoggy:
Looks promising !!!

PR #1824, opened 2017-12-28
Treat LFS files as large binaries in the web viewer

yaroot:
Currently in master, if you point the web ui to a file stored with LFS, it will dump the pointer file instead of the actual content (except for images). This PR makes the viewer treats LFS files as binaries, allow users to download them without seeing the gibberish text (does not affect images).

Another solution would be treating the LFS file the same as normal file, show the content if it's a text file.


takezoe:
@yaroot Thanks!

PR #1825, opened 2017-12-29
Replace H2 console

takezoe:
GitBucket includes H2 console as a database viewer. However, it has some problems like following:

- User interface isn't fit with GitBucket
- We have to fill connection information every time
- We can't add new features

So we will replace it with GitBucket native feature. I'm working on `features/dbviewer` branch and this is a current screenshot:


uli-heller:
Does this work with other dbs? For example mariadb?

takezoe:
@uli-heller Yes, of course!

PR #1830, opened 2018-01-04
use addSbtCoursier

xuwei-k:
- https://github.com/coursier/coursier/blob/v1.0.0/sbt-coursier/src/main/scala/coursier/CoursierPlugin.scala#L57-L59
- https://github.com/coursier/coursier/commit/a743f1183e6412c205b17d52f7bf1d6e3cd09161


takezoe:
Should we update `project/project/plugin.sbt` as well?

xuwei-k:
> Should we update project/project/plugin.sbt as well?

impossible. 

https://github.com/coursier/coursier/commit/a743f1183e6412c205b17d52f7bf1d6e3cd09161

> If sbt-coursier is enabled in `project/project/plugins.sbt`, then one
can just add `addSbtCoursier` in `project/plugins.sbt` instead of
`addSbtPlugin(…)`. In particular, that allow to specify the version of
coursier only once.

takezoe:
Thanks, I didn't understand this helper is to make possible to upgrade coursier by just updating `project/project/plugin.sbt`.

PR #1857, opened 2018-01-24
Return 401 for non-browsers when authentication is failed

takezoe:
The current version of GitBucket redirects to the sign-in page when authentication is failed. By this behavior, if we try to download a zip file from a repository using CLI tool like curl and authentication is failed, CLI gets HTML of the sign-in page with 200.  It confuses us because request appears to be successful.

This pull request makes the authentication error doesn't redirect for non-browser clients and just returns 401 to them.


McFoggy:
LGTM

I performed some quick tests using http://woothee.github.io/ and http://httpbin.org/user-agent with different browsers & CLI commands to see if woothee was able to find good categories or not ; this PR should work as expected.

PR #1878, opened 2018-02-05
Assign dummy password to newly created users

### LINKED ISSUE #1877: Openid Connect users ca not authenticate using git CLI client.
## Issue
**Impacted version**: 4.21.2 (but I suspect affects versions from 4.20 onwards)

**Deployment mode**: standalone app with an http frontend (apache httpd reverse proxy)

**Problem description**:
- Users created used OpenId Connect are not able to authenticate using the "git" command line.
  This means that users can't do `git clone` or `git clone`.
- This is because users created using "OpenId Connect" do not get assigned a password in the
  "ACCOUNT" database table (this is expected, as the authentication is done via OpenId).  This
  is a problem because if the "git" command tries to authenticate with username and password,
  gitbucket only checks the "ACCOUNT" database table.
- A possible solution would be, that when the account is created in [AccountFederationService](https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/service/AccountFederationService.scala#L37), line 37, instead of using an empty string for the password, a default
  possible random password is assigned.  That way, when the user sees their own profile
  they can change it to a known.  Then they can use this known password to authenticate.
- Using the database viewer, I was able to set a password in the Account's table.  Doing so
  allowed me to use the git CLI.

alejandroliu:
Closes #1877 

Set's the passwords of users created through OpenId connect to a dummy value.  This allows the user
to reset their passwords at their convenience.  This then let's them use "git" command line to
authenticate push/pull requests. 

### DIFF
--- a/src/main/scala/gitbucket/core/service/AccountFederationService.scala
+++ b/src/main/scala/gitbucket/core/service/AccountFederationService.scala
@@ -34,7 +34,7 @@ trait AccountFederationService {
         None
       case None =>
         findAvailableUserName(preferredUserName, mailAddress) flatMap { userName =>
-          createAccount(userName, "", fullName.getOrElse(userName), mailAddress, isAdmin = false, None, None)
+          createAccount(userName, "JUNKJUNK", fullName.getOrElse(userName), mailAddress, isAdmin = false, None, None)
           createAccountFederation(issuer, subject, userName)
           getAccountByUserName(userName)
         }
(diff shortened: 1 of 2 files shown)


takezoe:
@alejandroliu Test fails: https://travis-ci.org/gitbucket/gitbucket/builds/337799247

alejandroliu:
Hello,
All tests pass now.

int128:
This may not safe because any user can sign in with the same password.

alejandroliu:
User's can not sign in with the same password.  Although `createAccount` is used with "[DUMMY]" hard-coded as a password, this is still safe, because  normally the password would be the result of a "sha1" calculation of the clear-text password. So trying to login with "[DUMMY]" as a password will *always* fail.
BTW, the text "[DUMMY]" is also a string that would never be generated by the "sha1" function.

int128:
Thanks for the explanation. I got it.

PR #1896, opened 2018-03-05
Simple css change.

stheno:
Keep everything nice and tidy in database viewer without having to scroll the whole page.

### DIFF
--- a/src/main/webapp/assets/common/css/gitbucket.css
+++ b/src/main/webapp/assets/common/css/gitbucket.css
@@ -128,6 +128,12 @@ div.content-wrapper {
   display: block;
   position: relative;
   overflow: scroll;
+  max-height: 400px;
+}
+
+#table-tree {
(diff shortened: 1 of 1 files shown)


stheno:
Optimized for a 1980x1050 screen, but it hard caps the max scroll of entire screen to a smaller value.

takezoe:
I tested this modification, but I don't agree with limiting height of the tree viewer and the table viewer.

stheno:
Fair enough. I'll use it local then. I found it kind of obnoxious the multiple scrollings that happen and how it looks as a result. Thought that others my see the same.

aadrian:
@stheno there should be only one scroll, if the content is bigger than the screen (otherwise it should use the entire available area).

SIkebe:
>Fair enough. I'll use it local then.

4.34.0 introduced user-defined css (#2458). So this type of customization can be easy.

PR #1917, opened 2018-03-20
Add plugin list API

kounoike:
I want this feature to use in plugin building farm.

### DIFF
--- a/src/main/scala/gitbucket/core/api/ApiPlugin.scala
+++ b/src/main/scala/gitbucket/core/api/ApiPlugin.scala
@@ -1,0 +1,17 @@
+package gitbucket.core.api
+
+import gitbucket.core.plugin.{PluginRegistry, PluginInfo}
+
+case class ApiPlugin(
+  id: String,
+  name: String,
+  versino: String,
+  description: String,
+  jarFileName: String
+)
+
+object ApiPlugin{
+  def apply(plugin: PluginInfo): ApiPlugin = {
+    ApiPlugin(plugin.pluginId, plugin.pluginName, plugin.pluginVersion, plugin.description, plugin.pluginJar.getName)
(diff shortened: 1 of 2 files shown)


takezoe:
API implementation looks good except that a path is not necessary to be under `/api/v3`. I think GitBucket specific APIs should be separated from GitHub compatible APIs such as `/api/gitbucket/plugins`.

> I want this feature to use in plugin building farm.

Could you tell how do you use this API  in plugin build farm?

kounoike:
> API implementation looks good except that a path is not necessary to be under /api/v3. I think GitBucket specific APIs should be separated from GitHub compatible APIs such as /api/gitbucket/plugins.

I think it is better too. but, such implementation needs larger impacts:

- `ApiController` mounted on "/api/v3"
- `apiAuthenticationFilter` registered on "/api/v3/*"
- and other changes?

Can I change these implementations?

> Could you tell how do you use this API in plugin build farm?

I want to check each plugin successfully loaded or not. In current version, it can be checked with System administration > Plugins page. But it requires sign in (it means, form control/session control required).

If this PR is merged, I can be checked by `curl | jq` or `curl | grep`.

kounoike:
How about `/api/v3/gitbucket/plugins`? It means "v3-compatible GitBucket specific API". So, maybe we will implement GraphQL based API (`/api/graphql`) in the future.

takezoe:
> I think it is better too. but, such implementation needs larger impacts:

Completely understandable. `/api/v3/gitbucket/plugins` seems to make sense.

> I want to check each plugin successfully loaded or not.

OK, I understood that we will able to check whether plugins are actually loadable on specific version of GitBucket using this API.

kounoike:
@takezoe I changed API URL.

PR #1942, opened 2018-04-06
Add X-Forwarded-Proto support for baseurl parsing

### LINKED ISSUE #1941: OIDC redirect URI does not respect x-forwarded-proto header
I am using GitBucket behind a SSL load balancer. I found an OpenID Connect redirect URI is `http` even if `x-forwarded-proto` header is set to `https`. It is strange that others such as login redirect are `https`.

We can fix a redirect URI to `https` by setting Base URL in the system settings for now, but it should be set by the `x-forwarded-proto` header.

## Issue
**Impacted version**: 4.23.0

**Deployment mode**: 
```
Web Browser
↓HTTPS
AWS ALB (with ACM SSL certiticate)
↓HTTP
nginx (Kubernetes Ingress Controller)
↓HTTP
GitBucket (Kubernetes pod)
```

The nginx is configured to passthrough `x-forwarded-*` headers.

yaroot:

close #1941

### DIFF
--- a/src/main/scala/gitbucket/core/service/SystemSettingsService.scala
+++ b/src/main/scala/gitbucket/core/service/SystemSettingsService.scala
@@ -1,7 +1,6 @@
 package gitbucket.core.service
 
 import javax.servlet.http.HttpServletRequest
-
 import com.nimbusds.jose.JWSAlgorithm
 import com.nimbusds.oauth2.sdk.auth.Secret
 import com.nimbusds.oauth2.sdk.id.{ClientID, Issuer}
@@ -155,6 +154,8 @@ trait SystemSettingsService {
 object SystemSettingsService {
   import scala.reflect.ClassTag
 
+  private val HttpProtocols = Vector("http", "https")
+
   case class SystemSettings(
     baseUrl: Option[String],
     information: Option[String],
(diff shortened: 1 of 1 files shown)


yaroot:
Need feedback from #1941

src/main/scala/gitbucket/core/service/SystemSettingsService.scala
```diff
@@ -177,11 +179,18 @@ object SystemSettingsService {
...
+
+    def parseBaseUrl(req: HttpServletRequest): String = {
+      val url = req.getRequestURL.toString
+      val path = req.getRequestURI
+      val contextPath = req.getContextPath
+      val len = url.length - path.length + contextPath.length
+
+      val base = url.substring(0, len).stripSuffix("/")
+      Option(req.getHeader(XForwardedProto))
+        .fold(base)(_.toLowerCase() + base.dropWhile(_ != ':'))
```
takezoe: I think it might be dangerous to believe `X-Forward-Proto` header unconditinally.
It safer to confirm whether the value of `X-Forward-Proto` is https, then replace protocol if it's true.

int128: FYI: Scalatra checks if the header is `https` as https://github.com/scalatra/scalatra/blob/be9fc5119e6e34c4882d848cf86d9808acea8b68/core/src/main/scala/org/scalatra/ScalatraBase.scala#L696-L700

yaroot: @int128 as far as I can see, here we can not use Scalatra's built-in stuff. I'd rather be wrong here because I want to use framework support to handle this instead of rolling my own solution.

src/main/scala/gitbucket/core/service/SystemSettingsService.scala
```diff
@@ -1,14 +1,14 @@
...
 import javax.servlet.http.HttpServletRequest
-
 import com.nimbusds.jose.JWSAlgorithm
 import com.nimbusds.oauth2.sdk.auth.Secret
 import com.nimbusds.oauth2.sdk.id.{ClientID, Issuer}
 import gitbucket.core.service.SystemSettingsService._
 import gitbucket.core.util.ConfigUtil._
 import gitbucket.core.util.Directory._
 import gitbucket.core.util.SyntaxSugars._
+import org.eclipse.jetty.http.{HttpHeader, HttpScheme}
```
takezoe: These classes are not available at runtime. It's compilable, but I run GitBucket by `sbt ~jetty:start` or run the executable war file, I got `java.lang.NoClassDefFoundError: org/eclipse/jetty/http/HttpHeader`.

yaroot: fixed

takezoe:
> is there akka-http or http4s equivalent Request/Response type in scalatra (that's better typed and handles protocol related stuffs)?

Unfortunately, Scalatra doesn't have such types. It rather decorates servlet API using implicit conversion.

As you know, Scalatra isn't a modern framework in the Scala world any more. Scalatra team had a plan to adopt http4s as its backend in the past, but I guess it's difficult to be realized in the near future. So moving to raw http4s (or Akka HTTP) from Scalatra might be more realistic solution for GitBucket.

PR #1964, opened 2018-04-20
sort username for issue comments

kounoike:
Sort usernames for useful order.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/IndexController.scala
+++ b/src/main/scala/gitbucket/core/controller/IndexController.scala
@@ -33,6 +33,7 @@ trait IndexControllerBase extends ControllerBase {
   self: RepositoryService
     with ActivityService
     with AccountService
+    with IssuesService
     with RepositorySearchService
     with UsersAuthenticator
     with ReferrerAuthenticator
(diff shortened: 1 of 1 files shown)


src/main/scala/gitbucket/core/controller/IndexController.scala
```diff
@@ -194,18 +195,62 @@ trait IndexControllerBase extends ControllerBase {
...
+        val userCommentMap = comments.groupBy(c => c.commentedUserName).mapValues { lst =>
+          lst.map { c =>
+            commentIdMax - c.commentId
+          }.max
+        }
+        val groupMembers = getGroupMembers(owner).map(_.userName)
+        val issue = getIssue(owner, repo, num).get
+        a: Account =>
+          {
+            val typeStr = if (Some(a.userName) == context.loginAccount.map(_.userName)) {
```
takezoe: `context.loginAccount.contains(a.userName)` is much simpler.

src/main/scala/gitbucket/core/controller/IndexController.scala
```diff
@@ -194,18 +195,62 @@ trait IndexControllerBase extends ControllerBase {
     contentType = formats("json")
     val user = params("user").toBoolean
     val group = params("group").toBoolean
+
+    val issueRe = (context.baseUrl + """/(\w+)/(\w+)/(?:issues|pull)/(\d+)""").r
+    val sortByFunc: Account => String = context.request.referrer match {
```
takezoe: Getting repository and issue (or pull request) identifier from referrer is nice idea, but I think it's little tricky. It might be better to pass these information as optional query string parameters.

kounoike: As I know, this request is called from `helper/attached.scala.html` through `helper/preview.scala.html`.  `preview.scala.html` called from issue/pulls comment, wiki edit and release message edit. So it is complicated to add optional query string.

takezoe: Ah... Okay. It wasn't so easy. Anyway, I don't prefer kind of ad-hoc solution. I will try to find an another way. Sorry for keeping this pull request reviewing.

PR #1977, opened 2018-04-26
[WIP]Fix column type and validation for long webhook URL

### LINKED ISSUE #1951: AWS CodeBuild requires a very long webhook URL and token
I am trying GitBucket and AWS CodeBuild for continuous integration. AWS CodeBuild can generate a webhook URL and token but they are too long and GitBucket does not accept them.

For example, in the us-west-2 zone, a webhook URL seems 280~290 chars as like:

```
https://codebuild.us-west-2.amazonaws.com/webhooks?t=[VERY LONG STRING]&v=1
```

and a token seems 1000~1100 chars. As far as I know, AWS does not provide max length of webhook URL and token, but I think the followings are enough:

- Webhook URL: 1024 chars
- Security token: 2048 chars

Currently the webhook form accepts a very long string because it does not have length validation, but finally it causes a database error.

## Issue
**Impacted version**: 4.23.0

**Deployment mode**: Running in a container behind a reverse proxy and SSL load balancer on AWS

int128:
This fixes #1951.

### DIFF
--- a/src/main/resources/update/gitbucket-core_4.24.xml
+++ b/src/main/resources/update/gitbucket-core_4.24.xml
@@ -7,4 +7,7 @@
 
   <addPrimaryKey constraintName="IDX_ACCOUNT_EXTRA_MAIL_ADDRESS_PK" tableName="ACCOUNT_EXTRA_MAIL_ADDRESS" columnNames="USER_NAME, EXTRA_MAIL_ADDRESS"/>
   <addUniqueConstraint constraintName="IDX_ACCOUNT_EXTRA_MAIL_ADDRESS_1" tableName="ACCOUNT_EXTRA_MAIL_ADDRESS" columnNames="EXTRA_MAIL_ADDRESS"/>
+
+  <modifyDataType tableName="WEB_HOOK" columnName="URL" newDataType="text"/>
+  <modifyDataType tableName="WEB_HOOK_EVENT" columnName="URL" newDataType="text"/>
 </changeSet>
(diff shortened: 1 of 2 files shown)


int128:
I will check that MySQL, PostgreSQL and H2 supports column change to TEXT from VARCHAR. Please merge after that.

int128:
Oops. It needs an index for TEXT column but it is not supported. Any ideas?
A webhook URL should be longer than 255 and cannot be VARCHAR.

takezoe:
I have two ideas:

- Modify (or remove) constraints and keep consistency by application layer
- Add alternative columns which store hash value of text columns, and use these columns for constraints

I think the second one is a right way but it needs complex migration.

ms502040:
I exactly dont'n know about H2 and pgsql, but ***mysql*** max varchar is 65,535 and max row length 65,535 bytes too [The CHAR and VARCHAR Types](https://dev.mysql.com/doc/refman/8.0/en/char.html), [Row Size Limits](https://dev.mysql.com/doc/refman/8.0/en/column-count-limit.html):   
**H2** [Limits and Limitations](http://www.h2database.com/html/advanced.html#limits_limitations)
>There is **no limit** for the following entities, except the memory and storage capacity: maximum identifier length (table name, column name, and so on); maximum number of tables, columns, indexes, triggers, and other database objects; maximum statement length, number of parameters per statement, tables per statement, expressions in order by, group by, having, and so on; maximum rows per query; maximum columns per table, columns per index, indexes per table, lob columns per table, and so on; maximum row length, index row length, select row length; **maximum length of a varchar column**, decimal column, literal in a statement.

**pgsql** [Character Types](https://www.postgresql.org/docs/9.2/static/datatype-character.html):
> In any case, the longest possible character string that can be stored is about 1 GB

PR #1990, opened 2018-04-29
Add absolute date/time display option

### LINKED ISSUE #816: 'x' days ago is not a useful information in commit list.
Maybe it looks nice in "recent activity" but in the commit history, the "'x' days ago" information is not useful at all. Please put the timestamp there.

See in the screenshot below:
gitbucket_commit_time

how not helpful at all is the information **"2 days ago"** and **"3 days ago"**. More practical would have been to have the time, since the commits are grouped by date. Of course, after the date, one could extend with the text of how much time is that date in the past, but for single commits the timestamp is more helpful.

This is of course how GitHub works, but IMHO GitBucket should not "clone" the bugs too :).

Thank you.

kounoike:
This PR can close #816. Add "Absolute/Relative" time option to system settings. By default, it still uses relateive time format.


src/main/twirl/gitbucket/core/helper/datetimeago.scala.html
```diff
@@ -1,5 +1,5 @@
 @(latestUpdatedDate: java.util.Date,
-  recentOnly: Boolean = true)
+  recentOnly: Boolean = true)(implicit context: gitbucket.core.controller.Context)
```
takezoe: It might be not necessary to provide this as Twirl template. How about moving this to helper?

src/main/scala/gitbucket/core/view/helpers.scala
```diff
@@ -34,29 +34,37 @@ object helpers extends AvatarImageProvider with LinkConverter with RequestCache
...
    * Format java.util.Date to "x {seconds/minutes/hours/days/months/years} ago"
    */
-  def datetimeAgo(date: Date): String = {
-    val duration = new Date().getTime - date.getTime
-    timeUnits.find(tuple => duration / tuple._1 > 0) match {
-      case Some((unitValue, unitString)) =>
-        val value = duration / unitValue
-        s"${value} ${unitString}${if (value > 1) "s" else ""} ago"
-      case None => "just now"
+  def datetimeAgo(date: Date)(implicit context: Context): String = {
```
takezoe: It's nice to apply that the result of this method (and also `datetimeAgoRecentOnly`) is changed by `context.settings.relativeTime` to Scaladoc comment.

src/main/twirl/gitbucket/core/helper/datetimeago.scala.html
kounoike: It means "move to `gitbucket.core.view.helpers`?"

takezoe: Yes, but I think we might have to consider about behavior of date formatting again because it has little complex behavior currently. For example, `datetimeAgoRecentOnly` returns absolute date for old dates in current implementation.

In addition, I think `Date.toString()` is not good solution to generate display strings.

PR #1995, opened 2018-05-04
Introduce new Renderer

kounoike:
# Introduce **new** rendering approach

This PR introduce `EnhancedRenderer` and some extended classes.

### All files rendered by `EnhancedRenderer`.

- It makes more simple logic for blob/diff view.
- Previous `Renderer` assumes used for render *Markup*. So, image/video/audio renderer is tricky.

### correct handling image files

Previously, GitBucket detects image files by MIME-type. But, some `image/xxx` files can't render by browser. for example tiff file can't render by `img` tag.

`EnhancedRenderer` provides `ImageRenderer` trait. It represents "file can render by img tag".

### Image file is not limited to binary

For example, SVG file is text but it can rendered by img tag. So this PR introduce `TextImageRenderer`. It extends `TextRenderer` and `ImageRenderer`. For `TextImageRenderer` repository viewer and diff viewer render both image and source text .

image

image

### Allow read file as string or binary for `EnhancedRenderer`

`EnhancedRenderer` can read file as binary. So, it can render with format converter (for example, tiff to png).

It enables direct serving image file with data-URI. It reduce number of connections.

### Diffs rendered by `EnhancedRenderer`.

If plugin has new renderer which extends `ImageRenderer`, automatically rendered with Image Diff tool.

or, If renderer extends `DiffRenderer`, diffs are rendered with custom diff rendering.

for example: JSON diff by [benjamine/JsonDiffPatch](https://github.com/benjamine/JsonDiffPatch)

image

## Compatibility

The `RendererWrapper` wraps old-style `Renderer` to new-style `EnhancedRenderer`. So, no changes required for old-style Renderer plugins.

### DIFF
--- a/src/main/scala/gitbucket/core/plugin/EnhancedRenderer.scala
+++ b/src/main/scala/gitbucket/core/plugin/EnhancedRenderer.scala
@@ -1,0 +1,165 @@
+package gitbucket.core.plugin
+
+import java.util.Base64
+
+import gitbucket.core.controller.Context
+import gitbucket.core.service.RepositoryService
+import gitbucket.core.util.JGitUtil.ContentInfo
+import gitbucket.core.util.StringUtil
+import gitbucket.core.view.helpers
+import gitbucket.core.view.helpers.urlLink
+import org.apache.http.util.TextUtils
+import play.twirl.api.Html
+
+import scala.collection.immutable
+
+/**
+ * New style Renderer
+ */
+trait EnhancedRenderer {
+  def renderContent(request: EnhancedRenderRequest): Html = {
+    this match {
+      case r: TextRenderer =>
+        r.renderAsText(request)
+      case r: ImageRenderer =>
+        r.renderAsImage(request, "")
+      case r: ObjectRenderer =>
+        r.renderAsObject(request)
+      case _ =>
+        Html("??????")
+    }
+  }
+
+  val allowLargeFile: Boolean = false
+}
+
+trait TextRenderer extends EnhancedRenderer {
+  def renderAsText(request: EnhancedRenderRequest): Html
+  def renderMarkup(request: RenderRequest): Html
+}
+
+trait ImageRenderer extends EnhancedRenderer {
+  def renderAsImage(request: EnhancedRenderRequest, attr: String): Html
+}
+
+trait TextImageRenderer extends TextRenderer with ImageRenderer {
+  override def renderContent(request: EnhancedRenderRequest): Html = {
+    new Html(immutable.Seq[Html](renderAsImage(request, ""), renderAsText(request)))
+  }
+}
+
+trait ObjectRenderer extends EnhancedRenderer {
+  def renderAsObject(request: EnhancedRenderRequest): Html
+}
+
+trait DiffRenderer extends EnhancedRenderer {
+  def renderDiff(oldRequest: EnhancedRenderRequest, newRequest: EnhancedRenderRequest): Html
+}
+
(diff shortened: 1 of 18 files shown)


kounoike:
@takezoe Thanks for your review comments. Sorry for large PR. But I can't split to smaller PR.

I fixed this PR with your comments except [this comment](https://github.com/gitbucket/gitbucket/pull/1995#discussion_r187768852). It is essential part of this PR. I think we need to discuss it carefully. I think `renderAsXXX` is easily understandable for plugin developers. Especially its target file format can render as both text and image, like SVG.

src/main/scala/gitbucket/core/util/JGitUtil.scala
```diff
@@ -136,32 +144,177 @@ object JGitUtil {
...
    */
-  case class ContentInfo(viewType: String, size: Option[Long], content: Option[String], charset: Option[String]) {
+  case class ContentInfo(
+    owner: String,
+    repository: String,
+    commitId: String,
+    loader: ObjectLoader,
+    branch: String,
+    filePath: String
+  ) {
```
takezoe: Basically, case classes shouldn't cause (or depend on) side effect in the constructor. These properties should be set at the outside of case classes (e.g. You can create a factory method in the companion object to do it).

This also makes possible to extract a renderer using pattern matching, not cast of the renderer property.

src/main/scala/gitbucket/core/plugin/EnhancedRenderer.scala
```diff
@@ -0,0 +1,165 @@
...
+import gitbucket.core.view.helpers.urlLink
+import org.apache.http.util.TextUtils
+import play.twirl.api.Html
+
+import scala.collection.immutable
+
+/**
+ * New style Renderer
+ */
+trait EnhancedRenderer {
```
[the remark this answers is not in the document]
takezoe: Well... My opinion is the a parent class shouldn't know sub classes.

`renderContent()` isn't necessary to be implemented at `EnhancedRenderer`. This method should be only declared at `EnhancedRenderer`, and you can give implementation at `TextRemderer`, `ImageRenderer` and `ObjectRenderer`.

src/main/twirl/gitbucket/core/helper/diff.scala.html
```diff
@@ -101,34 +103,112 @@
...
-              @if(oldCommitId.isDefined && diff.oldIsImage){
-                <div class="diff-image-frame diff-old"><img src="@helpers.url(repository)/raw/@oldCommitId.get/@diff.oldPath" class="diff-image" onload="onLoadedDiffImages(this)" style="display:none" /></div>
-              } else {
-                @if(diff.changeType != ChangeType.ADD){
-                  <div style="padding: 12px;">Not supported</div>
+          @(diff.oldContentInfo, diff.newContentInfo) match {
+            case (None, None) => {
+              <div>Error???</div>
+            }
+            case (Some(oldContentInfo: ContentInfo), Some(newContentInfo: ContentInfo)) if oldContentInfo.renderer.isInstanceOf[DiffRenderer] && newContentInfo.renderer.isInstanceOf[DiffRenderer] => {
```
takezoe: I wonder if renderer can be extracted from `ContentInfo` using pattern matching, not type casting. See also above commet for `ContentInfo`.

takezoe:
@kounoike I added some comment to your new rederer framework. I hope that you will check them.

ExtArtQ:
And how did it all end?

PR #1997, opened 2018-05-06
[WIP] Experimental use webpack

kounoike:
I tried to introduce webpack. This PR contains the result.

## advantage to current implementation

- resource bundling

Now, many resources are included from every page. Webpack can bundle these resources.

before:

after:

Of cource, these resources are cached by browser normally.

- minification

In webpack and sbt-webpack, minify js/css is simply done by `webpack prod` on sbt.

## advantage to Webjars (#1595)

- Webjars is hard-tied to sbt and node.js

Rather, I think that it is appropriate to say that "Webjars is overkill". It imports npm dependency to jars one, so more lines required to write in build.sbt.

Webjars also has more complicated dependency by webjars classic/npm/bower jars. At least I know, trouble will occur when using mixed them.

- management library

webpack uses simple npm's `package.json`. It can use GitHub's repo directory. So, we can manage our customized `jsdifflib` as gitbucket/jsdifflib.

- simplicity

If we needs new javascript library, adding it is very easy. `npm install --save` and `require('library');` to `main.js`.

## disadvantage

- need node.js/npm install for developer

**User** doesn't need node/npm. Simply run `java -jar gitbucket.war`. But for building GitBucket, `npm install` downloads node_modules and sbt-webpack requires node.js.

## TODOs

I didn't check all functionality behavior. At least, diff doesn't work yet. It is required to do modulerize our javascript files.

### DIFF
--- a/webpack.config.base.js
+++ b/webpack.config.base.js
@@ -59,7 +59,7 @@ module.exports = {
       fonts: [
         { family: "Source Sans Pro"}
       ],
-      outputDir: "target/webapp/assets/fonts"
+      path: "target/webapp/assets/fonts"
     })
   ]
 };


nazoking:
I changed my mind. I think webpack is a good solution for gitbucket. More than webjars.

.gitignore
```diff
@@ -24,3 +24,6 @@ project/plugins/project/
 .idea/
 .idea_modules/
 *.iml
+/package-lock.json
```
nazoking: https://docs.npmjs.com/files/package-lock.json
> This file is intended to be committed into source repositories,

aadrian:
As discussed here https://github.com/gitbucket/gitbucket/issues/1430 too, webpack is the better solution.

zexxar:
I also think that webpack is the better solution.

PR #2013, opened 2018-05-19
Change to avatar image link from avatar image

kazuki43zoo:
I will suggest to add a link on avatar image at possible places.


takezoe:
@kazuki43zoo By these changes, `avatar()` looks to be used only in `avatarLink()`. So I think it can be private or inlined, but we might have to take care of use in plugins.

kazuki43zoo:
I use the avatar at some places yet because it's not uses the link at username part.

takezoe:
I see. So it makes sense to keep avatar method.

PR #2003, opened 2018-05-12
View badges on pull request screen

kazuki43zoo:
### Before submitting a pull-request to GitBucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [x] rebased my branch over master
- [x] verified that project is compiling
- [x] verified that tests are passing
- [x] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [ ] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

### Description

I've added following badges on pull request screen.

* conversation size
* commit size
* changed file size

These badges are exists until the GitBucket 4.23.x. But from 4.24.x these badges has been removed.
I want to revert viewing these badges.

WDYT?

PR #2015, opened 2018-05-19
Counting conversations on pull request screen

kazuki43zoo:
In current implementation, conversation count become all comment size. In other words, it includes action comments(e.g. `add label`, `remove label`, and so on...).
I will expect to display only comments for commit comment and pull request comment. (I think action of PR does not need on conversation count)

### Related Issues

* #2003

### DIFF
--- a/src/main/twirl/gitbucket/core/pulls/menu.scala.html
+++ b/src/main/twirl/gitbucket/core/pulls/menu.scala.html
@@ -8,6 +8,8 @@
   repository: gitbucket.core.service.RepositoryService.RepositoryInfo,
   flash: Map[String, String] = Map.empty)(body: => Html)(implicit context: gitbucket.core.controller.Context)
 @import gitbucket.core.view.helpers
+@import gitbucket.core.model.Comment
+@import gitbucket.core.model.CommitComments
 @import gitbucket.core.model.IssueComment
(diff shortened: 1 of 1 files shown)


src/main/scala/gitbucket/core/view/helpers.scala
```diff
@@ -463,6 +463,16 @@ object helpers extends AvatarImageProvider with LinkConverter with RequestCache
...
 
+  /**
+   * Count conversations(commit-comment or issue-comment action).
+   *
+   * @param comments all comments(includes actions)
+   * @return conversation count
+   */
+  def countConversation(comments: Seq[Comment]): Int = {
+    comments.count(c => c.isInstanceOf[CommitComments] || c.asInstanceOf[IssueComment].action.endsWith("comment"))
+  }
```
takezoe: I'm not sure whether we should add this method to helpers because this isn't a generic method for view construction. What do you think about following alternatives:

- Write inline this counting or define the method in the view
- Count in the controller and pass it to the view

kazuki43zoo: @takezoe Thanks for your review!
I've changed to count conversations at the view of pull request screen.

takezoe:
@kazuki43zoo Thank you!

Issue #118, opened 2013-10-04
Password security


phxql:
Do not use SHA-1 for password hashing, it's insecure. Use PBKDF2 or bcrypt for secure password storage.

PR #2017, opened 2018-05-19
use PBKDF2 for password

kounoike:
This PR improves password hashing discussed in #118. Newly created account or changed password are stored with PBKDF2 with SHA256. Old style password is also checked by `defaultAuthentication`. And it automatically update to PBKDF2.

### DIFF
--- a/src/main/scala/gitbucket/core/service/AccountService.scala
+++ b/src/main/scala/gitbucket/core/service/AccountService.scala
@@ -38,7 +38,7 @@ trait AccountService {
       case account if !account.isGroupAccount =>
         account.password match {
           case pbkdf2re(iter, salt, hash) if (pbkdf2_sha256(iter.toInt, salt, password) == hash) => Some(account)
-          case p: String if p == sha1(password) =>
+          case p if p == sha1(password) =>
             updateAccount(account.copy(password = pbkdf2_sha256(password)))
             Some(account)
           case _ => None
--- a/src/main/scala/gitbucket/core/service/AccountService.scala
+++ b/src/main/scala/gitbucket/core/service/AccountService.scala
@@ -33,7 +33,16 @@ trait AccountService {
    * Authenticate by internal database.
    */
   private def defaultAuthentication(userName: String, password: String)(implicit s: Session) = {
+    val pbkdf2re = """^\$pbkdf2-sha256\$(\d+)\$([0-9a-zA-Z+/=]+)\$([0-9a-zA-Z+/=]+)$""".r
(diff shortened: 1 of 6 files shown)


src/main/scala/gitbucket/core/service/AccountService.scala
```diff
@@ -33,7 +33,16 @@ trait AccountService {
    * Authenticate by internal database.
    */
   private def defaultAuthentication(userName: String, password: String)(implicit s: Session) = {
+    val pbkdf2re = """^\$pbkdf2-sha256\$(\d+)\$([0-9a-zA-Z+/=]+)\$([0-9a-zA-Z+/=]+)$""".r
     getAccountByUserName(userName).collect {
+      case account if !account.isGroupAccount =>
+        account.password match {
+          case pbkdf2re(iter, salt, hash) if (pbkdf2_sha256(iter.toInt, salt, password) == hash) => Some(account)
+          case p: String if p == sha1(password) =>
```
takezoe: Is type pattern `: String` necessary here?

takezoe:
@kounoike I added tiny comments. In particular, I would like to hear your opinion about transforming existing passwords.

kounoike:
@takezoe GitBucket doesn't store plain password. So, we can't transform it automatically except authentication phase.

mgeorge:
I am concerned that this change will kill the only mechanism that we 
have for batch creating accounts.  We can currently do this directly via 
the database console or the database import function.  We can easily 
generate SHA1 hashes using available command-line tools.

Since the API does not yet support account creation we have no other 
option for batch creation.

n 23/05/18 18:59, Naoki Takezoe wrote:

yaroot:
@mgeorge sha1 will not be removed, newly created account (via gitbucket) will have pbkdf2 as the hash function. This is a backward compatible change.

takezoe:
@kounoike 
>  GitBucket doesn't store plain password.

Hehe, you are right. Never mind.

src/main/scala/gitbucket/core/service/AccountService.scala
takezoe: I mean it's rewritable as:
```scala
case p if p == sha1(password) =>
```

takezoe:
@kounoike A comment is remaining. Could you check here?
https://github.com/gitbucket/gitbucket/pull/2017/files#diff-8627cf516a3a5c11771924bf62d4b4cbR41

kounoike:
@takezoe fixed!

PR #2014, opened 2018-05-19
improve archive download

kounoike:
This PR contains two archive improvements.

1. It can download LFS contents correctly (it can close #1998).
2. It can download subtree of repository.

### Before submitting a pull-request to GitBucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [ ] rebased my branch over master
- [x] verified that project is compiling
- [x] verified that tests are passing
- [ ] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [ ] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

PR #2027, opened 2018-05-23
Fix for repository downloading improvement

takezoe:
- Replace a custom exception with `NoSuchElementException`
- Remove unnecessary code

@kounoike This is tiny fix for #2014. Could you review it?


takezoe:
In addition, I think it might be better that a download file contains the directory name like when downloding the sub tree of the repository.

PR #2083, opened 2018-06-29
Drop JRebel configuration

takezoe:
Because myJRebel will be terminated at July, 5, 2018. We will became to not able to use it without purchasing for the license.


aadrian:
@takezoe what about http://hotswapagent.org/ ?

takezoe:
Thanks. I've never heared that. Is there someone who has tried that?

Issue #2048, opened 2018-06-03
Broken archive download


t3rkwd:
## Issue
**Impacted version**: 4.25.0

**Deployment mode**: standalone app

**Problem description**:
- Can not download the tar.gz file

```
12:39:45.645 [qtp522764626-16] ERROR g.c.c.RepositoryViewerController - Catch unhandled error in request: Request(GET //localhost:8080/root/gitbucket/archive/4.25.0.tar.gz)@589c541f
java.lang.NullPointerException: null
	at org.eclipse.jgit.lib.ObjectIdOwnerMap.get(ObjectIdOwnerMap.java:138)
	at org.eclipse.jgit.revwalk.RevWalk.parseAny(RevWalk.java:887)
	at gitbucket.core.util.JGitUtil$.getRevCommitFromId(JGitUtil.scala:219)
	at gitbucket.core.controller.RepositoryViewerControllerBase.$anonfun$archiveRepository$1(RepositoryViewerController.scala:1148)
	at gitbucket.core.controller.RepositoryViewerControllerBase.$anonfun$archiveRepository$1$adapted(RepositoryViewerController.scala:1146)
	at gitbucket.core.util.SyntaxSugars$.using(SyntaxSugars.scala:40)
	at gitbucket.core.controller.RepositoryViewerControllerBase.archive$1(RepositoryViewerController.scala:1146)
	at gitbucket.core.controller.RepositoryViewerControllerBase.$anonfun$archiveRepository$9(RepositoryViewerController.scala:1216)
	at gitbucket.core.controller.RepositoryViewerControllerBase.$anonfun$archiveRepository$9$adapted(RepositoryViewerController.scala:1212)
	at gitbucket.core.util.SyntaxSugars$.using(SyntaxSugars.scala:15)
	at gitbucket.core.controller.RepositoryViewerControllerBase.$anonfun$archiveRepository$8(RepositoryViewerController.scala:1212)
	at gitbucket.core.controller.RepositoryViewerControllerBase.$anonfun$archiveRepository$8$adapted(RepositoryViewerController.scala:1211)
	at gitbucket.core.util.SyntaxSugars$.using(SyntaxSugars.scala:15)
	at gitbucket.core.controller.RepositoryViewerControllerBase.archiveRepository(RepositoryViewerController.scala:1211)
	at gitbucket.core.controller.RepositoryViewerControllerBase.archiveRepository$(RepositoryViewerController.scala:1137)
[...]

PR #2102, opened 2018-07-11
Fix for broken archive downloads (see the issue #2048)

pazaur:
Fix for archive download behavior both for "Download Zip" button and for releases download (tar.gz, bz2 etc. ).
- Corrected URL to the "githublike".
	was: archive/master/master.zip (archive/branch/name/branch/name/name.zip)
	is: archive/master.zip (archive/branch/name.zip)
	The output file still is reponame-branch-name.zip format though.
- Complex named (branch/foo/bar etc.) branches are correctly compressed and downloaded.
	Earlier it was empty zip file and NPE in logs.

See the issue #2048 (closed but not fixed apparently) and comments within it.

affected versions: 4.25, 4.26

Tested it as much as I could. :)

### DIFF
--- a/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
+++ b/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
@@ -836,7 +836,7 @@ trait RepositoryViewerControllerBase extends ControllerBase {
    */
   get("/:owner/:repository/archive/:branch.zip")(referrersOnly { repository =>
     val branch = params("branch")
-    archiveRepository(branch, branch + ".zip", repository, "")
+    archiveRepository(branch, branch, repository, "", ".zip")
   })
 
   /**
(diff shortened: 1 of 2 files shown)


takezoe:
@pazaur The current implementation intends to archive a certain directory when "Download ZIP" button is clicked at a sub directory. This pull request seems to ignore this intention. It seems to archive a entire repository even if the button is clicked not at the root of a repository.

RadWolfie:
@takezoe 

My understanding GitHub perform the archive download as this method.
`http://[host]/[user]/[repo]/archive/[branch].zip`
Plus it doesn't have sub directory support.

If we really need sub directory support, then gitbucket can't use exact query like GitHub do. Plus require completely different type of url query. To something like...
`http://[host]/[user]/[repo]/archive/[branch]/[directory].zip`
which is identical to (functional in gitbucket)
`http://[host]/[user]/[repo]/tree/[branch]/[directory]`

Keep in mind, git's branch creation do support backslashes.

takezoe:
See #2104

takezoe:
Fixed in #2104

PR #2104, opened 2018-07-15
Fix archiving for slashed branches

takezoe:
This pull request fixes archiving for branches which contain `/` in their name by url encoding a branch name in following URL form:

- `baseUrl`/archive/`branch`.zip

When archiving sub directories, URL form is:

- `baseUrl`/archive/`path`/`branch`.zip

See also #2048 and #2102

### DIFF
--- a/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
+++ b/src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
@@ -834,52 +834,37 @@ trait RepositoryViewerControllerBase extends ControllerBase {
   /**
    * Download repository contents as a zip archive as compatible URL.
    */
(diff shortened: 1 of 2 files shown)


pazaur:
@takezoe Now, when I try to download archive from a "release (tag)" page (the Source Code link there). 
http://localhost:8080/root/test/releases 
link: http://localhost:8080/root/test/archive/test-tag.zip (where "test-tag" is a tag name)
I get:

HTTP ERROR 404
Problem accessing /root/test/archive/test-tag.zip. Reason:
    Not Found

takezoe:
@pazaur Oops, I'll fix it immediately. Thanks for reporting!

takezoe:
Additionally fixed in 33a676f221f3c9b6bd59e1f2c986533bbe316794

PR #2127, opened 2018-08-04
Display fullname

beckorz:

FullNameを表示できるように修正しました。


takezoe:
I think the necessity of displaying the full name is depends on the situation. For example, we need to know the user name when we want to mention on the issues or pull requests. So displaying the user name is useful in this situation.

I think there is no obvious answer about this. So I would like to hear your opinion.

PaulAdams-Conti:
In our company we only have uids as user names (from Active Directory), so it is very difficult to assign real people to issues or see exactly who has committed (if they have not configured their .gitconfig correctly). So having the possibility to see the full names instead of uidxxxxx would be useful in all situations for us.

takezoe:
Hm, I understood that displaying full name is useful in some cases. Making user name presentation configurable sounds good. @beckorz closed this pull request, but I wonder if we can start to implement that based on this pull request.

kounoike:
I found old issue #336

YukiMori81:
Sorry for digging up old pull request.
In our company, username is simple unique six digit number, so it's so difficult to recognize who had committed (I can know the real person name if I click the username).
I think it's helpful for some users if LDAP displayname is shown in the News feed instead of username. So if it can be configurable in the system settings, it would be very helpful.

Thank you for your time.

PR #2187, opened 2018-11-05
add openjdk 8 & 11 in the travis-ci build matrix

McFoggy:
as discussed on gitter, introduce opendjdk in travis build to ensure compatibility with openjdk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: true
 jdk: 
   - oraclejdk8
   - oraclejdk11
+  - openjdk8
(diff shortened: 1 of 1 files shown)


takezoe:
Should we keep OracleJDK?

McFoggy:
I would go for keeping it. The new licenses program does not apply to development and it ensures that gitbucket passes the tests under those JVMs.
Then it is up to each of the consumer to choose their runtime JVM depending on their needs/budget.

takezoe:
I see. Keep them for now.

PR #2202, opened 2018-11-26
Basic implementation of OpenId Connect groups

aliuly:
Hi,
This PR is more for a call for comments than anything else.  This PR lets you synchronize group membership from an OpenID Connect provider that provides a "groups" claim.

I am curious if other people are interested in something like this.  This PR is based on 4.29.0.  If other people are interested, then I could create a PR based on the master branch for contributing to the project.

To use it with KeyCloak, you need to add a Mapper to your client definition with the following:

- Name: groups
- Mapper Type: Group Membership
- Token Claim Name: groups
- Full group path: false
- Rest set to default

When OIDC is used, this path only gets activated if the passed Token contains a "groups" claim.  If it does, then it will look for group names in the "groups" claim named "Group-admin" or "Group-user".  So for example, if the Token groups claim contains a group named "Contoso-admin", then, the user will be added to the "Contoso" group as manager.  On the other hand, if the groups claim contains a group named "Contoso-user", then the user will be added to the "Contoso" group as a normal member.

Simlarly, if the user belongs in a group  and the corresponding "group-user" or "group-admin" is not found in the Token "groups" claim, the user will be removed from the group.

Note that groups must already exist in gitbucket, for this to work.  This means that if the "groups" claim contain multiple groups not used by gitbucket, these will be ignored.

This lets you manage users from an external Identity Provider system and control group membership from there.  Which makes the tool more suitable for Enterprise users.

### DIFF
--- a/src/main/scala/gitbucket/core/service/AccountService.scala
+++ b/src/main/scala/gitbucket/core/service/AccountService.scala
@@ -282,7 +282,49 @@ trait AccountService {
     List(userName) ++
       Collaborators.filter(_.collaboratorName === userName.bind).sortBy(_.userName).map(_.userName).list.distinct
   }
-
+  def isUserGroupMember(
+    groupName: String,
+    userName: String
+  )(implicit s: Session) : List[Boolean] = {
+    GroupMembers
+      .filter(_.userName === userName.bind)
+      .filter(_.groupName === groupName.bind)
+      .map(_.isManager)
+      .list
+  }
+  def removeUserFromGroup(
+    groupName: String,
+    userName: String
+  )(implicit s: Session) : Unit = {
+    GroupMembers
+      .filter(_.userName === userName.bind)
+      .filter(_.groupName === groupName.bind)
+      .delete
+  }
+  def addUserToGroup(
+    groupName: String,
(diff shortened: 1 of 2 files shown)


aliuly:
The travis check failed (wrong formatting).  I will fix it (if there is interest for this PR).

PS: Also, apologies, I don't really know how to program in Scala.

Grishnackh:
This is basically what I requested in issue #2316.

BlueIcarus:
@aliuly I know your pull-request is a little old but have you made any changes to the code to allow the pull-request to merge? I would really appreciate this functionality as its a hassle to manage the permissions using the GitBucket web interface, API or direct database changes and I would prefer this to be automatic (as it would be with your changes).

aliuly:
@BlueIcarus Unfortunately I am not working on GitBucket anymore.

We were using it as pat of a project at work, but that project has since been cancelled.

alwibrm:
@takezoe Any chance we get this merged? :)

takezoe:
Of course, yes if it's mergeable. However, I don't have enough time and motivation to fix and test this pull request. I wish someone review and test this pull request instead.

PR #2207, opened 2018-12-04
Allow login form to accept username in case-insensitive manner

XD-DENG:
## Issue that this PR is trying to solve

Currently the username while login is case-sensitive. For example, using "ROOT" to login will fail while "root" will do.
screen shot 2018-12-04 at 9 06 13 pm

However, creating new user with ID "ROOT" will fail, which means the username should be case-insensitive.
screen shot 2018-12-04 at 9 08 19 pm

## Benefit brought by this PR

**This PR makes the login form accept username with case-insensitive manner. Say my ID is `test`, then entering `test` or `TEST` or `Test` in login form should all be valid.**

**This is also how most applications handles case of ID (including GitHub).**

### DIFF
--- a/src/main/scala/gitbucket/core/controller/IndexController.scala
+++ b/src/main/scala/gitbucket/core/controller/IndexController.scala
@@ -88,7 +88,7 @@ trait IndexControllerBase extends ControllerBase {
   }
 
   post("/signin", signinForm) { form =>
-    authenticate(context.settings, form.userName, form.password) match {
+    authenticate(context.settings, form.userName.toLowerCase, form.password) match {
       case Some(account) =>
         flash.get(Keys.Flash.Redirect) match {
           case Some(redirectUrl: String) => signin(account, redirectUrl + form.hash.getOrElse(""))


aadrian:
-1 for this.

Usernames should be case sensitive. Dependening on the configured login, this is also passed further to other systems that are also case sensitive.

XD-DENG:
Like the example I gave earlier, then should creating `root` and `ROOT` as two different users allowed?

To me, the answer is no. "root" and "ROOT" is the same ID. Then user should be allowed to use either one to login.

XD-DENG:
Thanks @aadrian for sharing your thoughts. 

1. For ID creation, being case-insensitive can help avoid conflict thoroughly (which is also the current practice of GitBucket I think).

2. OS platform/cross platform should not be a concern, because the ID is always stored/matched in lower-case eventually. The only change is to allow users to enter their ID in either upper or lower case. As mentioned above, this is also the practice of many applications/websites.

But eventually this is up to you and other maintainers :-)

aadrian:
> For ID creation, being case-insensitive can help avoid conflict thoroughly (which is also the current practice of GitBucket I think).

@XD-DENG as I mention in previous comment, usernames **can't** be case insenstive, since the username and the password need to be passed further to other systems: LDAP, or Oauth, so a login in this case would not work at all, so forcing this assumption for everybody will break the existing systems.

This is why usernames should **always** be case senstive from the user's point of view in any well designed system. Not allowing other users with the same name (in a case insesntive matter) is a different matter, and is a preffered solution to avoid further conflicts.

Also note that the username is part of the URL, and URLs by defintion are case insenstive too.

takezoe:
There is a historical background in the case sensitivity of username in GitBucket.

As @aadrian said, allowing case sensitive username is for the integration with external systems and it's also for the backward compatibility. Since we had to keep both them, adding case insensitive validation in user registration on GitBucket was the only thing we could do to prevent further registration of case-different usernames.

XD-DENG:
Thanks @takezoe for sharing the background.

Please feel free to close this PR if it can not be considered.

BTW GitBucket is a very well-done job. Thanks.

PR #2232, opened 2018-12-24
[WIP] Add tests for WebHook models JSON serialization

takezoe:

### DIFF
--- a/src/test/scala/gitbucket/core/service/WebHookJsonFormatSpec.scala
+++ b/src/test/scala/gitbucket/core/service/WebHookJsonFormatSpec.scala
@@ -1,0 +1,38 @@
+package gitbucket.core.service
+
+import org.json4s.jackson.JsonMethods.parse
+import org.json4s._
+import org.scalatest.FunSuite
+
+class WebHookJsonFormatSpec extends FunSuite {
+  import gitbucket.core.api.ApiSpecModels._
+
+  test("WebHookCreatePayload"){
+    fail("TODO")
+  }
+
+  test("WebHookPushPayload"){
+    fail("TODO")
+  }
+
+  test("WebHookIssuesPayload"){
+    fail("TODO")
+  }
+
+  test("WebHookPullRequestPayload"){
+    fail("TODO")
+  }
+
+  test("WebHookIssueCommentPayload"){
+    fail("TODO")
+  }
+
+  test("WebHookPullRequestReviewCommentPayload"){
+    fail("TODO")
+  }
+
+  test("WebHookGollumPayload"){
+    fail("TODO")
+  }
+
+}


src/test/scala/gitbucket/core/service/WebHookJsonFormatSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+                          |"watchers_count":0,
+                          |"url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World",
+                          |"http_url":"http://gitbucket.exmple.com/git/octocat/Hello-World.git",
+                          |"clone_url":"http://gitbucket.exmple.com/git/octocat/Hello-World.git",
+                          |"html_url":"http://gitbucket.exmple.com/octocat/Hello-World"
+                          |}""".stripMargin
+
+  private def assert(payload: WebHookPayload, expected: String): Assertion = {
+    val json = JsonFormat(payload)
+    assert(json == expected.replaceAll("\n", ""))
```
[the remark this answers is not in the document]
shimamoto: Though I thought so, what do you think?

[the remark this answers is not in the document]
takezoe: That makes sense. We should rewrite existing tests to not use json4s functionallity for assertion even though it would be really painful.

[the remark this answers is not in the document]
shimamoto: Will do. I'll handle it on this PR.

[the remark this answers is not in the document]
takezoe: Great! Thanks!!

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
```diff
@@ -42,228 +48,609 @@ object ApiSpecModels {
...
+       |"number":1347,
+       |"title":"new-feature",
+       |"user":$jsonUser,
+       |"labels":[$jsonLabel],
+       |"state":"closed",
+       |"created_at":"2011-04-14T16:00:49Z",
+       |"updated_at":"2011-04-14T16:00:49Z",
+       |"body":"Please pull these awesome changes",
+       |"id":0,
+       |"comments_url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/issues/1347/comments",
```
shimamoto: Is this `comments_url` correct?

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
```diff
@@ -42,228 +48,609 @@ object ApiSpecModels {
...
+       |"user":$jsonUser,
+       |"labels":[$jsonLabel],
+       |"assignee":$jsonUser,
+       |"id":0,
+       |"html_url":"http://gitbucket.exmple.com/octocat/Hello-World/pull/1347",
+       |"url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/pulls/1347",
+       |"commits_url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/pulls/1347/commits",
+       |"review_comments_url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/pulls/1347/comments",
+       |"review_comment_url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/pulls/comments/{number}",
+       |"comments_url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/issues/1347/comments",
```
shimamoto: Is this `comments_url` correct?

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
```diff
@@ -42,228 +48,609 @@ object ApiSpecModels {
...
+       |"committer":{"name":"octocat","email":"octocat@example.com","date":"2011-04-14T16:00:49Z"},
+       |"url":"http://gitbucket.exmple.com/octocat/Hello-World/commit/$id"
+       |}""".stripMargin
+
+  // TODO comment_url
+  val jsonCommits = s"""{
+       |"url":"http://gitbucket.exmple.com/api/v3/repos/octocat/Hello-World/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e",
+       |"sha":"6dcb09b5b57875f334f61aebed695e2e4193db5e",
+       |"html_url":"http://gitbucket.exmple.comoctocat/Hello-World/commit/6dcb09b5b57875f334f61aebed695e2e4193db5e",
+       |"comment_url":"http://gitbucket.exmple.com",
```
shimamoto: Is this `comment_url` correct?

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
```diff
@@ -42,228 +48,609 @@ object ApiSpecModels {
...
+  val jsonCommit = (id: String) => s"""{
+       |"id":"$id",
+       |"message":"full message",
+       |"timestamp":"2011-04-14T16:00:49Z",
+       |"added":[],
+       |"removed":[],
+       |"modified":["README.md"],
+       |"author":{"name":"octocat","email":"octocat@example.com","date":"2011-04-14T16:00:49Z"},
+       |"committer":{"name":"octocat","email":"octocat@example.com","date":"2011-04-14T16:00:49Z"},
+       |"url":"http://gitbucket.exmple.com/octocat/Hello-World/commit/$id"
```
shimamoto: `url` doesn't include path `api/v3/repos`. Is this correct?

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
takezoe: It seems to that we don't have API for commit comment. We might have to remove this property from the model.

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
takezoe: It's correct in `WebHookJsonFormatSpec`, but not correct in `JsonFormatSpec`.

In this case, `apiCommit` is created as follows:
```scala
  val apiCommit = {
    val commit = commitInfo(sha1)
    ApiCommit(
      id = commit.id,
      message = commit.fullMessage,
      timestamp = commit.commitTime,
      added = Nil,
      removed = Nil,
      modified = List("README.md"),
      author = ApiPersonIdent.author(commit),
      committer = ApiPersonIdent.committer(commit)
    )(repo1Name, true)
  }
```
The last parameter (true) is `urlIsHtmlUrl`. This parameter should be `true` when this model is used as web hook payload, and should be `false` when this mode is used as response of API.

So, ideally we might have to define two models. One is for `JsonFormatSpec` created with `urlIsHtmlUrl = false` and another one is for `WebHookJsonFormatSpec` created with `urlIsHtmlUrl = true`.

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
takezoe: Probably this is correct. I suppose that since the comments API for pull request seems to be not implemented yet, this property points API for the issues instead.

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
takezoe: Same as above.

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
shimamoto: ok

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
shimamoto: ok

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
shimamoto: ok

src/test/scala/gitbucket/core/api/ApiSpecModels.scala
shimamoto: https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/api/ApiCommit.scala#L26-L35

Only this switches output attributes, and the others output both `url` and `html_url`. GitHub API also looks like it outputs both.

PR #2243, opened 2019-01-06
Add ETag feature for cache plugin-assets

kounoike:
In #1491, I added Cache-Control header for cache plugin-assets. But, it is over-cached for plugin-developers. It requires reload when plugin updated.

I replaced cache control with ETag header which generated by plugin jar's modified time.

### DIFF
--- a/src/main/scala/gitbucket/core/plugin/PluginRegistry.scala
+++ b/src/main/scala/gitbucket/core/plugin/PluginRegistry.scala
@@ -409,6 +409,13 @@ object PluginRegistry {
     }
   }
 
+  def getPluginInfoFromClassLoader(classLoader: ClassLoader): Option[PluginInfo] = {
+    instance
+      .getPlugins()
+      .find { info =>
+        info.classLoader.equals(classLoader)
+      }
+  }
 }
 
 case class Link(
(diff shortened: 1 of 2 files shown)


src/main/scala/gitbucket/core/plugin/PluginRegistry.scala
```diff
@@ -409,6 +409,13 @@ object PluginRegistry {
     }
   }
 
+  def getPluginInfoFromClassLoader(classLoader: ClassLoader): Option[PluginInfo] = {
```
takezoe: Using ClassLoader as key of PluginInfo is an idea. Ideally, we might have to make possible to provide a general way to get PluginInfo by the plugin id in the future. But it would require large refactoring across the plugin system.

Anyway, `getPluginInfoFromClassLoader` should not be at a general place as a public API. I think it should be implemented as a private or protected method in `PluginAssetsServlet`.

kounoike: > it should be implemented as a private or protected method in `PluginAssetsServlet`.

Can I get `instance` from `PluginAssetsServlet`?

takezoe: I mean this method shouln't be a public API. Being package private or any other solutions are acceptable. Don't you have good idea?

takezoe: I see. My idea is moving plugin related servlets to under `plugin` package, but your choise looks better.

PR #2290, opened 2019-03-27
switch to xhub4j-core 1.1.0 for java 9/11 compatibility

McFoggy:
xhub4j-core-1.1.0 that removes jaxb dependency is on its way to maven central.

it is already pushed to sonatype (1), next synchronization should occure within some hours. 
Once it is there, this PR could be merged to update the dependency.

1: https://oss.sonatype.org/content/groups/public/fr/brouillard/oss/security/xhub/xhub4j-core/1.1.0/


McFoggy:
it looks like the check status does not reflect the green travis build

takezoe:
Thanks!

PR #2326, opened 2019-06-05
Pubkey vs. Private Key clarity

burnsjake:
It should be clearer to they user that they're adding a public key for access to a git repo and not adding private keys unnecessarily.

### DIFF
--- a/src/main/twirl/gitbucket/core/account/ssh.scala.html
+++ b/src/main/twirl/gitbucket/core/account/ssh.scala.html
@@ -1,9 +1,9 @@
 @(account: gitbucket.core.model.Account, sshKeys: List[gitbucket.core.model.SshKey])(implicit context: gitbucket.core.controller.Context)
 @import gitbucket.core.ssh.SshUtil
-@gitbucket.core.html.main("SSH Keys"){
+@gitbucket.core.html.main("Public SSH Keys"){
   @gitbucket.core.account.html.menu("ssh", context.loginAccount.get.userName, false){
     <div class="panel panel-default">
-      <div class="panel-heading strong">SSH Keys</div>
+      <div class="panel-heading strong">Public SSH Keys</div>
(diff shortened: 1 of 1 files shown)


src/main/twirl/gitbucket/core/account/ssh.scala.html
```diff
@@ -19,7 +19,7 @@
     </div>
     <form method="POST" action="@context.path/@account.userName/_ssh" validate="true" autocomplete="off">
       <div class="panel panel-default">
-        <div class="panel-heading strong">Add a SSH Key</div>
+        <div class="panel-heading strong">Add an SSH public key</div>
```
McFoggy: wouldn't `Add a publis SSH key` be more suitable

src/main/twirl/gitbucket/core/account/ssh.scala.html
```diff
@@ -1,9 +1,9 @@
 @(account: gitbucket.core.model.Account, sshKeys: List[gitbucket.core.model.SshKey])(implicit context: gitbucket.core.controller.Context)
 @import gitbucket.core.ssh.SshUtil
-@gitbucket.core.html.main("SSH Keys"){
+@gitbucket.core.html.main("Public SSH Keys"){
```
McFoggy: I would keep the left menu as short as possible ; thus not changing it.

takezoe: @McFoggy This is used only for the title of HTML, so this change would be acceptable.

src/main/twirl/gitbucket/core/account/ssh.scala.html
takezoe:
```suggestion
        <div class="panel-heading strong">Add a SSH public key</div>
```
As @McFoggy says, this would be better.

Issue #1892, opened 2018-02-27
Too tight file size limitation in Release page (it's 3MB)


gageas:
Release page accepts only 3MB by default.
3MB is reasonable limitation for attachment to issues, but it is almost impossible to distribute modern software binaries.
(We know that gitbucket.war is bigger than 50MB)

https://github.com/gitbucket/gitbucket/blob/02a8540875abf8aea26163673cfdb48954f5d531/src/main/twirl/gitbucket/core/releases/form.scala.html#L63

https://github.com/gitbucket/gitbucket/blob/f64d4843f328cbbc8884d309cb6802146d521f7f/src/main/scala/gitbucket/core/util/FileUtil.scala#L82

takezoe:
Agree. Some endpoints like uploading release files and GitLFS should accept larger files than others. We should make them configurable in another configuration items.

lavvy:
So how do one even post a binary file to the repo through command line. For instance uploading a build artifact from a ci to a repo or release. There is nowhere i could find it in the wiki
i have tried this github command from here but no success .

curl 'https://api.github.com/repos/owner/my_repo/releases/:id/assets?access_token=acces_token&name=foo.zip' --header 'Content-Type: application/zip' --upload-file test.zip -X POST

takezoe:
@lavvy Uploading release files API hasn't been supported yet. Could you create an another issue to request it?

SIkebe:
@gageas As per https://github.com/gitbucket/gitbucket/pull/2363#issuecomment-522266565 , the `MaxFileSize` is configurable as follows.
`java -jar gitbucket.war --max_file_size=<max-file-size>`

Your suggestion is to have separate limits between release page and others?

gageas:
@SIkebe 
> Your suggestion is to have separate limits between release page and others?

Yes, it is.
I don't want to allow large file to Issue page, but to Release page.

GitHub accepts ~2GB for Release page, and ~25MB(~10MB for images) for Issue page.

skullwritter:
if @takezoe want's it, i may try to put it on the release as i have in my pull request for UploadTimeout
what u think @SIkebe

SIkebe:
>GitHub accepts ~2GB for Release page, and ~25MB(~10MB for images) for Issue page.

Thanks! @gageas I also found this.
https://help.github.com/en/articles/about-releases#limitations-on-binary-files
https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests

[...]

PR #2371, opened 2019-08-25
additional params to upload timeout and file size - Issue #1892

skullwritter:
This is relative to #1892


skullwritter:
@takezoe see if this is what you wanted as the params

SIkebe:
@skullwritter Could you react to a comment from @takezoe in this PR?
https://github.com/gitbucket/gitbucket/issues/1892#issuecomment-524649720

>How do you think about adding parameters for large files? Maybe, parameter names can be shortened though.
>
>- `--max_file_size_for_large_files`
>- `--upload_timeout_for_large_files`
>
>We can use these parameters for large file uploading. It might be only Releases for now, though.

Also it would be helpful if you edit the PR title more appropriate one.

skullwritter:
@SIkebe i've already reacted to him, thats why i've addered this pull request right after
and the title is corrected

skullwritter:
@SIkebe also, i've adhered the params as follows (in the comments from #1892 )
> @takezoe the params available on the new pull request are:
> 
> * `--max_file_size_issues=[MAX_FILE_SIZE_FOR_ISSUES]` (defaults to max_file_size)
> * `--upload_timeout_issues=[MAX_UPLOAD_TIME_IN_SECONDS_FOR_ISSUES]` (defaults to upload_timeout)
> * `--max_file_size_releases=[MAX_FILE_SIZE_FOR_RELEASES]` (defaults to max_file_size)
> * `--upload_timeout_releases=[MAX_UPLOAD_TIME_IN_SECONDS_FOR_RELEASES]` (defaults to upload_timeout)

SIkebe:
>i've already reacted to him

Yep, partly though. He suggested other param names. 
* `--max_file_size_for_large_files`
* `--upload_timeout_for_large_files`

skullwritter:
@SIkebe i can modify the params, the problem with that params in question is that they don't indicate where they will be applied

skullwritter:
@SIkebe  and the params as is, that would require rewriting the dropfile jquery plugin

skullwritter:
@SIkebe and adding other params like:
min file size to know if is a large file

SIkebe:
>the problem with that params in question is that they don't indicate where they will be applied

I think doc can cover that. In order to understand which size is allowed exactly, users have to read the doc. Param name doesn't have to cover all information.

GitBucket has several pages that allow uploading files; e.g. Issues, Pull requests, Releases, Wiki...
In addition, there will probably be more in the future.
Having limitation for each place is not very realistic.

So two limits, one for the default and the other one for large files, would be a viable option.

skullwritter:
hmmm..... the issues params i think it covers the pull request too, if i'm not mistaken, because i think that the issues and the pull requests uses the same form

takezoe:
Sorry for my very late reaction. In my understanding, remaining problem in this problem is only the parameter names. My current opinion is using general parameter names like:

- `--max_file_size_for_large_files`
- `--upload_timeout_for_large_files`

and describe features which these parameters are effective in README should be fine.

@SIkebe @skullwritter What do you think?

SIkebe:
@takezoe I am in favor of your proposal :-)

My opinion has not changed since I mentioned before.
>So two limits, one for the default and the other one for large files, would be a viable option.

skullwritter:
but what is the minimum size for a large file?

takezoe:
Is the minimum size limitation necessary?

skullwritter:
the minimum size to be considered a large file, yes, because, for every person the opinion differs...

takezoe:
I thought the max file size and timeout limitation will be applied to all file uploading at GitLFS, Issues, Pull requests, etc. There is no necessity to distinguish large files.

skullwritter:
> I thought the max file size and timeout limitation will be applied to all file uploading at GitLFS, Issues, Pull requests, etc. There is no necessity to distinguish large files.

we already have that, now for the
``--max_file_size_for_large_files``
``--upload_timeout_for_large_files``
we don't actually have a value for a file that is considered large

takezoe:
@skullwritter Hmm? What's your opinion? My proposal is that adding the following two options instead of per feature options (e.g. `max_file_size_issues` or `max_file_size_releases`)and they are applied to all file upload features and GitLFS.

- `--max_file_size_for_large_files`
- `--upload_timeout_for_large_files`

Does this make sense to you?

takezoe:
Also, we need to fix `FileUploadController` because it has the max file size limitation in `MultipartConfig`:
https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/controller/FileUploadController.scala#L33

mmestucci:
I saw the implementation about File Upload Timeout / Timeout for large files, but I still have an exception "java.util.concurrent.TimeoutException: Idle timeout expired: 3000/3000 ms" when I push a large lfs commit. It is like if this param is independent of these params. Is there other params to be configured?

PR #2395, opened 2019-11-04
Fix account table dead lock caused by ApiAuthenticationFilter

### LINKED ISSUE #2392: Database Deadlock when accessing /api/v3/user
### Before submitting an issue to GitBucket I have first:

- [X] searched for similar already existing issue
- [X] read the documentation and [wiki](https://github.com/gitbucket/gitbucket/wiki)

## Issue
**Impacted version**: 4.31.0 And Above 

**Deployment mode**: Docker container using https://hub.docker.com/r/f99aq8ove/gitbucket/ and PostgreSQL as Database Backend

**Problem description**:
When using API calls, to /v3/user and a valid authentication, the database enters in a deadlock state.

When too many request are required, the HikariCP connection pool starves, and no more request are served.

I've traced the issue down to PR #2267 . 
What I've seen at pgAdmin, is that there's a query idle in transaction, and another waiting to perform an update. 

The state after an attempt is:
* 'idle in transaction': Select data from Account
* 'locked': Update account set lastAction....

At the moment, a way to prevent gitbucket from getting starved is to set a flag to postgre user, that adds a timeout for connections that are 'idle in transaction'.

## Steps to reproduce
* Start a postgresql instance
* Configure gitbucket to use postgresql as database
* make a query to api `curl http://localhost/api/v3/user -u username:password` **before any user logs in**
* The request should not complete, unless you kill the postgresql connection (pgadmin should display the issue on the dashboard).

Edit: Added @SIkebe annotation on "how to reproduce"

SIkebe:

-----
Closes #2392 

1. Setup empty DB.
1. Run `java -jar gitbucket.war` for the first time.
1. `ScalatraBootstrap` registers `TransactionFilter` and then `ApiAuthenticationFilter`.
1. Run `curl http://localhost:8080/api/v3/user -u root:root`.
1. `TransactionFilter` begins transaction.
1. The default *root* user's password is SHA1 hash, so GitBucket tries to update the password by PBKDF2.
https://github.com/gitbucket/gitbucket/blob/f35ecce3c7b08e9579408252f80674b5b4ae7c43/src/main/scala/gitbucket/core/servlet/ApiAuthenticationFilter.scala#L29
https://github.com/gitbucket/gitbucket/blob/f35ecce3c7b08e9579408252f80674b5b4ae7c43/src/main/scala/gitbucket/core/service/AccountService.scala#L36-L43
1. `ApiAuthenticationFilter` begins another transaction to update `lastLoginDate` and causes dead lock.
https://github.com/gitbucket/gitbucket/blob/f35ecce3c7b08e9579408252f80674b5b4ae7c43/src/main/scala/gitbucket/core/servlet/ApiAuthenticationFilter.scala#L40-L42

@kounoike @takezoe  I'm not familiar with scalatra and slick yet, is my guess correct?

### DIFF
--- a/src/main/scala/gitbucket/core/controller/api/ApiUserControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiUserControllerBase.scala
@@ -4,7 +4,6 @@ import gitbucket.core.controller.ControllerBase
 import gitbucket.core.service.{AccountService, RepositoryService}
 import gitbucket.core.util.{AdminAuthenticator, UsersAuthenticator}
 import gitbucket.core.util.Implicits._
-import gitbucket.core.util.StringUtil._
 import org.scalatra.NoContent
 
 trait ApiUserControllerBase extends ControllerBase {
@@ -71,7 +70,7 @@ trait ApiUserControllerBase extends ControllerBase {
     } yield {
       val user = createAccount(
         data.login,
-        pbkdf2_sha256(data.password),
+        data.password,
         data.fullName.getOrElse(data.login),
         data.email,
         data.isAdmin.getOrElse(false),
--- a/src/main/scala/gitbucket/core/controller/api/ApiUserControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiUserControllerBase.scala
@@ -4,6 +4,7 @@ import gitbucket.core.controller.ControllerBase
 import gitbucket.core.service.{AccountService, RepositoryService}
 import gitbucket.core.util.{AdminAuthenticator, UsersAuthenticator}
 import gitbucket.core.util.Implicits._
+import gitbucket.core.util.StringUtil._
 import org.scalatra.NoContent
 
 trait ApiUserControllerBase extends ControllerBase {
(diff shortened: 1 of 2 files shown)


SIkebe:
FYI @Sherpard

SIkebe:
Although it's not directly related to the issue, the *root*'s default password should be hashed by PBKDF2?

kounoike:
@SIkebe Please do not remove this lock. So, it is meaningful lock for this scenario:

- login with api
- call some action which calls webhook
- in webhook side, also call api

If without new transaction, Dead lock occurs in this scenario.

But, I didn't check password update.

SIkebe:
@kounoike Thank you for pointing that out. I'll revert my change.

Authentication process for LDAP user also updates `mailAddress` and `fullName` just like SHA1-hashed password.
https://github.com/gitbucket/gitbucket/blob/f35ecce3c7b08e9579408252f80674b5b4ae7c43/src/main/scala/gitbucket/core/service/AccountService.scala#L54-L66

How should be these transactions treated?

takezoe:
@SIkebe Sorry for my very late reply.  I would like to merge this pull request and I think the subject of this pull request should be Hash password created by Web API as the last commit.

SIkebe:
@takezoe Thank you for reviewing. Could you review https://github.com/SIkebe/gitbucket/pull/1 ?
If that fix is appropriate, I'll merge @Sherpard 's commit into my branch and create another PR for my last commit.

takezoe:
Yeah. I'm just looking that pull request.

takezoe:
Thanks! @SIkebe and @Sherpard!

PR #2423, opened 2020-01-19
Add start and due date to issue

kasancode:
Squash #2422

This PR add `start date`, `due date` to issue.
This PR will change GitBucket version number because of updating database schema.

If this PR is merged, I will implement gantt chart plugin supporting Future Task.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import com.typesafe.sbt.pgp.PgpKeys._
 
 val Organization = "io.github.gitbucket"
 val Name = "gitbucket"
-val GitBucketVersion = "4.33.0"
+val GitBucketVersion = "4.34.0"
 val ScalatraVersion = "2.7.0-RC1"
 val JettyVersion = "9.4.26.v20200117"
 val JgitVersion = "5.6.0.201912101111-r"
--- a/src/main/resources/update/gitbucket-core_4.34.xml
+++ b/src/main/resources/update/gitbucket-core_4.34.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<changeSet>
+  <addColumn tableName="ISSUE">
+    <column name="START_DATE" type="datetime" nullable="true"/>
+    <column name="DUE_DATE" type="datetime" nullable="true"/>
+  </addColumn>
+</changeSet>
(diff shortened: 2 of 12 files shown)


takezoe:
Thanks for your contribution as always!

Though this pull request itself looks good to me, I would like to be careful to add new attributes to issues because I want to keep GitBucket core simple as much as possible and offload additional functionalities to plugins. However, determining what functionalities should be implemented as core and what should be plugins is not easy stuff.

Regarding `due date`, the milestone already has it. Of course, I understand the milestone's due date is not enough in the real project at work. Having `start date`, `due date` and even something which represent the amount of work for the issue (e.g. `story point`) would be useful for the project management. Adding the capability to enable to define custom attributes might be an option. However, it will raise performance concern as we saw in Trac long before.

Anyway, I want to hear opinion from users about this topic. Do you think adding `start date` and `due date` is worth to add to issues as a core functionality? Also, do you think we should add further attributes to issues (and pull requests)?

skullwritter:
@takezoe it depends, right now, the version that i have on my server it's the fastest, it that param is needed, or it's in github, for example, yes, put it. 

> If this PR is merged, I will implement gantt chart plugin supporting Future Task.

If that plugin will be beneficial for several users, that's great and it should be put in the main, since it's only a param....

Jookia:
@takezoe GitBucket already includes the ability to assign things, set labels and priorities and track milestone times, so adding due dates lets you prioritize that way too and covers the use case of time-sensitive tasks.

As for adding more fields to core, maybe? It kind of depends on their complexity I suppose. I'd like to have a Bugzilla-type issue blocking system so I can say 'this issue requires fixing A, B, C and blocks E, F, G'

takezoe:
Thanks for your comments! I want to consider that a capability to add custom fields rather than adding some particular fields. Plugins will be able to leverage this capability to add required fields. As a result of this consideration, if we conclude that supporting custom fields is difficult, we will back to this idea again.

kasancode:
@takezoe 
I understand your concept to keep core system simple. I will follow your decision. 

At first, I tried to implement `start date` and `due date` in plug-in using `issue comment`, that has `date` and `text` field.
However, `start date` and `due date` are useful in some plug-in like Gantt-chart, Kanban, and Calender for example.
So, I thought, these attributes are more suitable for the core system.

zsmahi:
I think it's a great idea to have this feature; In my case I'm using gitbucket in my company with a lot of plugins (kanban, network plugin) and having this feature will be beneficial for my team 👍

zsmahi:
Hi , any news about this feature ? is is included as feature or plugin ?

zsmahi:
Hello @kasancode has this feature been merged ?

takezoe:
My plan was to add an ability to define custom fields to issues and pull requests instead of adding specific fields. However, this plan hasn't made any progress due to the lack of our development resources.

To be honest, considering our current poor development resources and the difficulty of implementing such a complex customizability, I started feeling that merging this might be more realistic.

takezoe:
https://github.com/gitbucket/gitbucket/pull/2423

PR #2421, opened 2020-01-17
Limit the visible repositories in the sidebar.

kasancode:
This PR adds an "All / Limited" option to the system settings, and the "Limited" option reduces the visible repositories in the sidebar.
If you have too many users and repositories, this option will be useful.

### Before submitting a pull-request to GitBucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [x] rebased my branch over master
- [x] verified that project is compiling
- [ ] verified that tests are passing
- [ ] squashed my commits as appropriate *(keep several commits if it is relevant to understand the PR)*
- [ ] [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct

PR #2427, opened 2020-02-15
Ignore Limited settings, when searching repositories.

kasancode:
This PR is to ignore Limited settings (#2421) , when searching repositories.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/IndexController.scala
+++ b/src/main/scala/gitbucket/core/controller/IndexController.scala
@@ -293,7 +293,13 @@ trait IndexControllerBase extends ControllerBase {
 
     val repositories = {
       context.settings.limitVisibleRepositories match {
-        case true  => getVisibleRepositories(context.loginAccount)
+        case true =>
+          getVisibleRepositories(
+            context.loginAccount,
+            None,
+            withoutPhysicalInfo = true,
+            limit = false
+          )
         case false => visibleRepositories
       }
     }.filter { repository =>
--- a/src/main/scala/gitbucket/core/controller/IndexController.scala
+++ b/src/main/scala/gitbucket/core/controller/IndexController.scala
@@ -290,10 +290,16 @@ trait IndexControllerBase extends ControllerBase {
         withoutPhysicalInfo = true,
         limit = context.settings.limitVisibleRepositories
       )
-    val repositories = visibleRepositories.filter { repository =>
+
+    val repositories = {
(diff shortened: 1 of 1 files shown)


src/main/scala/gitbucket/core/controller/IndexController.scala
```diff
@@ -290,10 +290,16 @@ trait IndexControllerBase extends ControllerBase {
         withoutPhysicalInfo = true,
         limit = context.settings.limitVisibleRepositories
       )
-    val repositories = visibleRepositories.filter { repository =>
+
+    val repositories = {
+      context.settings.limitVisibleRepositories match {
+        case true  => getVisibleRepositories(context.loginAccount)
```
takezoe: Your fix looks to always disable `limit` option. I think it's better to modify the initial call as follows rather than make another call after checking `limitVisibleRepositories` option:
```scala
    val visibleRepositories =
      getVisibleRepositories(
        context.loginAccount,
        None,
        withoutPhysicalInfo = true,
        limit = false // fix this line
      )
```

kasancode: This url end point is required two type repository list.

```scala
gitbucket.core.search.html.repositories(
  query, 
  repositories,       // search result : always unlimited
  visibleRepositories // repo list in side bar : according to settings
  )
```

then, following implementation is also OK. however this always queries the DB twice.

```scala
get("/search") {
  val query = params.getOrElse("query", "").trim.toLowerCase

  val visibleRepositories =
    getVisibleRepositories(
    context.loginAccount,
    None,
    withoutPhysicalInfo = true,
    limit = context.settings.limitVisibleRepositories
    )

  val repositories = getVisibleRepositories(context.loginAccount).filter { repository =>
    repository.name.toLowerCase.indexOf(query) >= 0 || repository.owner.toLowerCase.indexOf(query) >= 0
  }

  gitbucket.core.search.html.repositories(query, repositories, visibleRepositories)
}
```

so, I think, previous implement is better.

takezoe: Ah, `repositories` and `visibleRepositories`. Makes sense.

Even though, `withoutPhysicalInfo` should be `false` in the second search too?

kasancode: Yes, you right.
I changed `withoutPhysicalInfo` to `true`.

Issue #2361, opened 2019-08-15
Error duaring configurating with MariaDB 10.4.7: "Incorrect index name 'ACCESS_TOKEN_ID'"


ytakahashi1981:
## Issue (Japanese below.)

**Impacted version**: 4.32.0
**MariaDB Version**: 10.4.7
**Deployment mode**: with Jetty 9.4.8

**Problem description**:
I tried to migrate database from h2database to mysql(mariadb), followed https://github.com/gitbucket/gitbucket/wiki/External-database-configuration (also https://takezoe.hatenablog.com/entry/2016/07/16/025154).
But after rebooting jetty, the following stacktrace has shown and GitBucket doesn't work.

```
. 2019-08-15 11:51:59.217:WARN:oejw.WebAppContext:main: Failed startup of context o.e.j.w.WebAppContext@3159c4b8{/gitbucket,file:///tmp/jetty-0.0.0.0-8080-gitbucket.war-_gitbucket-any-198649221540653045.dir/webapp/,UNAVAILABLE}{/gitbucket.war}
java.sql.SQLSyntaxErrorException: (conn=54140) Incorrect index name 'ACCESS_TOKEN_ID'
	at org.mariadb.jdbc.internal.util.exceptions.ExceptionMapper.get(ExceptionMapper.java:242)
	at org.mariadb.jdbc.internal.util.exceptions.ExceptionMapper.getException(ExceptionMapper.java:171)
	at org.mariadb.jdbc.MariaDbStatement.executeExceptionEpilogue(MariaDbStatement.java:248)
	at org.mariadb.jdbc.ClientSidePreparedStatement.executeInternal(ClientSidePreparedStatement.java:230)
	at org.mariadb.jdbc.ClientSidePreparedStatement.execute(ClientSidePreparedStatement.java:157)
	at org.mariadb.jdbc.ClientSidePreparedStatement.executeUpdate(ClientSidePreparedStatement.java:192)
	at com.zaxxer.hikari.pool.ProxyPreparedStatement.executeUpdate(ProxyPreparedStatement.java:61)
	at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeUpdate(HikariProxyPreparedStatement.java)
	at io.github.gitbucket.solidbase.migration.MigrationUtils.updateDatabase(MigrationUtils.java:20)
	at io.github.gitbucket.solidbase.migration.LiquibaseMigration.migrate(LiquibaseMigration.java:93)
[...]

PR #2435, opened 2020-03-29
Delete AddPrimaryKey tag which is not required with not composed inde…

nico57c:
This PR is to fix liquibase execution with MariaDb which return a "Multi primary key" error (#2361)

Tested on : 
  - H2 database
  - MariaDb _( Ver 15.1 Distrib 10.4.12-MariaDB, for Linux (x86_64) using readline 5.1 )_

_Plugin gist has same error :_

https://github.com/gitbucket/gitbucket-gist-plugin/blob/master/src/main/resources/update/gitbucket-gist_2.0.xml

### DIFF
--- a/src/main/resources/update/gitbucket-core_4.0.xml
+++ b/src/main/resources/update/gitbucket-core_4.0.xml
@@ -60,13 +60,12 @@
   <!-- ACCESS_TOKEN -->
   <!--================================================================================================-->
   <createTable tableName="ACCESS_TOKEN">
-    <column name="ACCESS_TOKEN_ID" type="int" nullable="false" autoIncrement="true" unique="true"/>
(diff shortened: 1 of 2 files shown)


takezoe:
Thanks for your contribution. I tested on PosgtreSQL and worked well.

takezoe:
By the way, I guess many GitBucket plugins use the same stuff. So, we might have to update them as well. At least, we need to see the default bundled plugins.

The following default bundled plugins use database:

- https://github.com/gitbucket/gitbucket-notifications-plugin
- https://github.com/gitbucket/gitbucket-gitbucket-plugin
- https://github.com/gitbucket/gitbucket-pages-plugin

takezoe:
I checked above plugins and there was no combination of `unique="true"` and `<addPrimaryKey .../>`. So, it should be okay.

PR #2465, opened 2020-06-20
Support both /ref and /refs in Get Reference API

takezoe:
The path of Get Reference API is `/ref` properly, some version of GHE, however, it's `/refs` in some versions of GHE. For external tools compatibility (e.g. Jenkins integration, etc), we need to support both paths in GitBucket.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/api/ApiGitReferenceControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiGitReferenceControllerBase.scala
@@ -5,6 +5,7 @@ import gitbucket.core.util.Directory.getRepositoryDir
 import gitbucket.core.util.ReferrerAuthenticator
 import gitbucket.core.util.Implicits._
 import org.eclipse.jgit.api.Git
+import org.slf4j.LoggerFactory
 
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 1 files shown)


src/main/scala/gitbucket/core/controller/api/ApiGitReferenceControllerBase.scala
```diff
@@ -17,6 +17,15 @@ trait ApiGitReferenceControllerBase extends ControllerBase {
    * https://developer.github.com/v3/git/refs/#get-a-reference
    */
   get("/api/v3/repos/:owner/:repository/git/ref/*")(referrersOnly { repository =>
+    getRef()
+  })
+
+  // Some versions of GHE support this path
+  get("/api/v3/repos/:owner/:repository/git/refs/*")(referrersOnly { repository =>
```
SIkebe, line 24: nit: Would it better to log warning here? Something like `logger.warn("git/refs/ endpoint may not be compatible with GitHub API v3. Consider using git/ref/ endpoint instead")` ?

takezoe: Thanks! Added a warn log to /refs.

PR #2475, opened 2020-07-04
CSS style: Don't hide focus indicators

Jookia:
Hiding focus indicators is unhelpful for people navigating by keyboard.
Show them instead.


src/main/webapp/assets/vendors/AdminLTE-2.4.2/css/AdminLTE.css
```diff
@@ -172,13 +172,6 @@ h6,
 a {
   color: #337ab7;
 }
-a:hover,
-a:active,
-a:focus {
-  outline: none;
-  text-decoration: none;
-  color: #23527c;
-}
```
SIkebe, line 181: What for?

BTW, do not modify this css file directly as it is from third party library.  Anything you modified would be lost when AdminLTE upgraded.

SIkebe: > What for?
> 

Sorry, you clarified here
>Hiding focus indicators is unhelpful for people navigating by keyboard.
Show them instead.

SIkebe: Hmm, why  did you remove `color: #23527c;`? Screenshots before and after would help.

Jookia:
Should font color change during hovering or focus? I'll add it back and
check it later today.

src/main/webapp/assets/vendors/AdminLTE-2.4.2/css/AdminLTE.css
Jookia: How else do I modify this CSS? It seems that the way AdminLTE indicates focus is by slightly darkening links. I can't even see the difference. So this commit replaces it with classic browser focus stuff. You can compare it by tabbing through the GitBucket demo and https://git.xogium.me/

SIkebe:
Ah I see, this color change is not from AdminLTE but Bootstrap. So your deletion does not have any effect.

color
-----

This `#23527c` is applied.

image

Jookia:
Ah, that's good. My intention is purely to make it so the focus box shows around things when using keyboard navigation, so you can see where you're at in the page.

SIkebe:
If your link has underline in it, this PR makes it harder to read correctly. Could you also tweak padding?

underline

Jookia:
Sure, how?

Jookia:
Ok I managed to reproduce it on my machine. I can't change that without changing the font used in Gitbucket. Maybe I should add the underline back?

SIkebe:
>Edit: For what it's worth I can't get your padding behavior on my browser.

That's strange... which browser are you using? As I said before screenshots would help.

SIkebe:
>I can't change that without changing the font used in Gitbucket.

How did you verify that? You can override any third party css behavior if you can define your style in gitbucket.css.
**Note: Do not use below style definition as it is. It affects all `<a>` tag.**

image

Jookia:
On Sat, Jul 04, 2020 at 10:12:02PM -0700, SIkebe wrote:
> >I can't change that without changing the font used in Gitbucket.
> 
> How did you verify that? You can override any third party css behavior if you can define your style in gitbucket.css.
> **Note: Do not use below style definition as it is. It affects all `<a>` tag.**
> 
> image

I disabled the link's font CSS in the Firefox inspector and it showed a
better font with better spacing for the underscores.

PR #2509, opened 2020-08-13
Concider branch name with '/' when downloading source codes

### LINKED ISSUE #2442: zip file cannot download if branch name include '/' charactor
## Issue
zip file cannot download if branch name include '/' charactor
**Impacted version**: 4.33.0

**Deployment mode**: 
GitBucket : standalone app with an http frontend (tomcat8)

**Problem description**:
- *be as explicit has you can*
- *describe the problem and its symptoms*
- *explain how to reproduce*
1. create branch that branch name include '/' charactor
2. try to download from created branch
3. URL error display 
    branch name '/' charactor is replaced "%2" on page URL
    I guess that is root cause.

SIkebe:

Fixes #2442


src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
```diff
@@ -846,15 +846,19 @@ trait RepositoryViewerControllerBase extends ControllerBase {
...
+    val name = params.get("splat")
+    if (name.isDefined) {
+      archiveRepository(name.get, repository, "")
+    }
   })
 
-  get("/:owner/:repository/archive/*/:name")(referrersOnly { repository =>
-    val name = params("name")
-    val path = multiParams("splat").head
-    archiveRepository(name, repository, path)
```
SIkebe, line 857: When branch name includes `/`, this endpoint was called even if you want to download whole repo. This causes `NullPointerException` later.

SIkebe: Is there any way to distinguish these routes without renaming?

takezoe: I thought we might be able to use the same logic as `tree` if URL has the following structure, for example:

- http://localhost:8080/root/repo/archive/release/v1.0.0?format=zip
- http://localhost:8080/root/repo/archive/release/v1.0.0/dir1/dir2?format=zip

However, this is just an idea. Since URL needs to be consistent and stable as much as possible, we need to design new URL very carefully if we will change it.

SIkebe:
>I thought we might be able to use the same logic as tree if URL has the following structure, for example:
>
>* http://localhost:8080/root/repo/archive/release/v1.0.0?format=zip
>* http://localhost:8080/root/repo/archive/release/v1.0.0/dir1/dir2?format=zip

@takezoe I applied your suggestion.

>we need to design new URL very carefully if we will change it.

Yeah, I completely agree with you, so I described as `breaking changes`.

I'm wondering how GitBucket distinguishes branch name and path correctly. There are two cases to consider.
* You want to download `v1.0.0/dir1/dir2` as a zip file from `release` branch.
* You want to download `dir1/dir2` as a zip file from `release/v1.0.0` branch.

Does `splitPath` assume that the former case is too rare?

SIkebe:
Rebased and resolved conflicts.

takezoe:
Sorry for very late reply!

> Does splitPath assume that the former case is too rare?

Certainly, if the repository has both `release` and `release/v1.0.0`, these two cases cannot be distinguished. Actually, many of repository viewer endpoints have the same issue.

src/main/twirl/gitbucket/core/releases/release.scala.html
```diff
@@ -48,8 +48,8 @@ <h3>
                 <span class="label label-default">@helpers.readableSize(Some(asset.size))</span>
               </li>
             }
-            <li><a href="@helpers.url(repository)/archive/@{helpers.encodeRefName(release.tag)}.zip"><i class="octicon octicon-file-zip"></i>Source code (zip)</a></li>
-            <li><a href="@helpers.url(repository)/archive/@{helpers.encodeRefName(release.tag)}.tar.gz"><i class="octicon octicon-file-zip"></i>Source code (tar.gz)</a></li>
+            <li><a href="@helpers.url(repository)/archive/@{helpers.encodeRefName(release.tag)}?format=zip"><i class="octicon octicon-file-zip"></i>Source code (zip)</a></li>
+            <li><a href="@helpers.url(repository)/archive/@{helpers.encodeRefName(release.tag)}?format=tar.gz"><i class="octicon octicon-file-zip"></i>Source code (tar.gz)</a></li>
```
takezoe, line 52: Could you also update `zipball_url` and `tarball_url` added in https://github.com/gitbucket/gitbucket/pull/2544? 🙇

src/main/scala/gitbucket/core/controller/RepositoryViewerController.scala
```diff
@@ -848,15 +848,14 @@ trait RepositoryViewerControllerBase extends ControllerBase {
...
-    val name = params("name")
-    archiveRepository(name, repository, "")
-  })
-
-  get("/:owner/:repository/archive/*/:name")(referrersOnly { repository =>
-    val name = params("name")
-    val path = multiParams("splat").head
-    archiveRepository(name, repository, path)
+  get("/:owner/:repository/archive/*")(referrersOnly { repository =>
+    val format = params("format")
```
takezoe, line 852: How about offering the previous bahvior if `format` parameter is not available for the backward compatibility? Generating a waring long in that case would be fine.

takezoe (changes requested):
@SIkebe Thanks for fixing the pull request. While the new path strategy for archiving looks good to me, I left some trivial comments. Could you take a look?

PR #2516, opened 2020-08-18
Show userName as link only when a user is not removed

SIkebe:

### before
before

-----

### after
after

### DIFF
--- a/src/main/twirl/gitbucket/core/admin/userlist.scala.html
+++ b/src/main/twirl/gitbucket/core/admin/userlist.scala.html
@@ -27,7 +27,11 @@
             </div>
             <div class="strong">
               @helpers.avatarLink(account.userName, 20)
-              <a href="@helpers.url(account.userName)">@account.userName</a>
+              @if(account.isRemoved){
+                @account.userName
+              } else {
+                <a href="@helpers.url(account.userName)">@account.userName</a>
+              }
               @if(account.isGroupAccount){
                 (Group)
               } else {


aadrian:
@SIkebe IMHO if the user was removed, but there are commits and other elements referenced to it's name, **only** the admin should be still able to see them. 

The username link should be disabled for normal users, or for the admins when the users had no activity.

SIkebe:
@aadrian Hmm, I consolidate the look & feel in user management page and issue/PR page.
Here's the image of a comment written by a removed user.
image

SIkebe:
>The username link should be disabled for normal users

Only administrators can see user management page, so this PR would not affect normal users.

aadrian:
> Only administrators can see user management page, 

But not the user page is linking to ! The URLs are predictable, and if someone has a link while the user existed (e.g. bookmarked), than it could still access it.

takezoe:
> The username link should be disabled for normal users, or for the admins when the users had no activity.

At least, git history will be kept even after removal of GitBucket account. I don't think we should revise them.

Anyway, I think that is out of the scope of this pull request.

aadrian:
@takezoe 

> At least, git history will be kept even after removal of GitBucket account. I don't think we should revise them.

Of course not. 

But if a user has been disabled (and not purged), and had activity, only the admin should be able to see the "user" page with that activity.

> I think how to treat removed accounts is out of the scope of this pull request.

OK.

SIkebe:
As @aadrian said, more granular management of activities or a feature like GitLab's Audit Events would be useful, for sure.
https://docs.gitlab.com/ee/administration/audit_events.html

I just wanted to remove useless link in this PR.

takezoe:
Let's consider treatment of removed accounts as a dedicated topic.

PR #2520, opened 2020-08-22
Call ReceiveHook when pull request is merged

takezoe:


takezoe:
If branch protection and status checking requirement are enabled, this causes the following error.

Screen Shot 2020-12-06 at 22 00 32

Because branch protection is implemented as a commit hook. Of course, we can bypass the branch protection in this particular case. However, I'm worrying that the same situation can happen with other plugins that use a commit hook.

PR #2523, opened 2020-08-24
adding support for wiki page deletion recording

### LINKED ISSUE #2522: Record DELETE activity of wiki page
## Issue
**Impacted version**: all

Currently GitBucket reocrd activity of wiki only for CREATED and EDITED.
Because `ActivityService `has only `recordCreateWikiPageActivity `and `recordEditWikiPageActivity`. It would be nice if there is `recordDeleteWikiPageActivity `too and record page deletion.

jyoo980:

I don't think I have the permissions to mark issues as closed, so I'll leave the last box unchecked. This may resolve #2522


src/main/scala/gitbucket/core/model/WikiActivityActionInfo.scala
```diff
@@ -0,0 +1,82 @@
+package gitbucket.core.model
+
+import java.util.UUID
+
+import gitbucket.core.model.Profile.currentDate
+
+sealed trait WikiActivityActionInfo {
```
jyoo980, line 7: By moving all the logic related with activity information into this sealed trait, I think we can enforce a better separation of concerns; we no longer pollute the `ActivityService` with the details of log formation, since we now delegate this responsibility to this trait and its subtypes.

src/main/scala/gitbucket/core/service/ActivityService.scala
```diff
@@ -335,46 +335,8 @@ trait ActivityService {
...
-        activityUserName,
-        "edit_wiki",
-        s"[user:${activityUserName}] edited the [repo:${userName}/${repositoryName}] wiki",
-        Some(pageName + ":" + commitId),
-        currentDate,
-        UUID.randomUUID().toString
-      )
-    )
-  }
+  def recordWikiActivity(maybeActivityInfo: Option[WikiActivityActionInfo]): Unit =
```
jyoo980, line 338: Converting the two functions we had into this generic function which accepts a trait that has subtypes specific to each type of activity has negated the need for having a specific function for each activity.

src/main/scala/gitbucket/core/servlet/GitRepositoryServlet.scala
```diff
@@ -523,6 +512,51 @@ class WikiCommitHook(owner: String, repository: String, pusher: String, baseUrl:
...
+            commit.committerName,
+            fileName.dropRight(".md".length),
+            commit.id
+          )
+        )
+      case _ => None
+    }
+  }
+
+  private[this] def mapToAction(changeType: ChangeType): String = changeType match {
```
jyoo980, line 554: Just wanted to make sure that these are all the actions we currently want to support.

src/main/scala/gitbucket/core/servlet/GitRepositoryServlet.scala
```diff
@@ -470,34 +475,18 @@ class WikiCommitHook(owner: String, repository: String, pusher: String, baseUrl:
...
+                      val maybeActivityInfo = buildActionInfo(commit, action, fileName)
+                      recordWikiActivity(maybeActivityInfo)
                       (action, fileName, commit.id)
                   }
                 }
               }
 
               val pages = commits
-                .groupBy { case (action, fileName, commitId) => fileName }
+                .groupBy { case (_, fileName, _) => fileName }
```
jyoo980, line 489: This `.groupby` was not using the values of `action` or `commitId`

src/main/scala/gitbucket/core/model/WikiActivityActionInfo.scala
takezoe: Do you intend all activity log generation currently implemented in `ActivityService` should be migrated to this way? I think it's better to migrate all activity logging methods at once if we introduce this than applying to only wiki activity logging and that refactoring might have to be another pull request.

src/main/scala/gitbucket/core/service/ActivityService.scala
```diff
@@ -335,46 +335,8 @@ trait ActivityService {
...
+    maybeActivityInfo.fold(())(activityInfo => writeLog(activityInfo.toLogEntry))
```
takezoe, line 339: In this case, we can use `foreach` instead of `fold`.

src/main/scala/gitbucket/core/servlet/GitRepositoryServlet.scala
```diff
@@ -523,6 +512,51 @@ class WikiCommitHook(owner: String, repository: String, pusher: String, baseUrl:
     }
   }
 
+  private[this] def buildActionInfo(
+    commit: JGitUtil.CommitInfo,
+    action: String,
+    fileName: String
+  ): Option[WikiActivityActionInfo] = {
+    action match {
```
takezoe, line 520: We may be able to use `ChangeType` directly for this pattern matching instead of the mapped string.

jyoo980: I used the mapped string here because we already use it downstream as a parameter to another function call.

src/main/scala/gitbucket/core/model/WikiActivityActionInfo.scala
jyoo980: I think this method of delegating the log formatting to the data types themselves might be a better idea than having a lot of very specific functions in `ActivityService`. 

I'd be happy to perform the refactoring myself. Maybe I can do it in a PR separate from this one and merge it in first and we can play these changes on top of it?

src/main/scala/gitbucket/core/servlet/GitRepositoryServlet.scala
takezoe: Oh, I see. Thanks for clarification.

src/main/scala/gitbucket/core/model/WikiActivityActionInfo.scala
takezoe: > I'd be happy to perform the refactoring myself. Maybe I can do it in a PR separate from this one and merge it in first and we can play these changes on top of it?

That sounds great.

> I tried to create an issue for this work but it looks like I don't have the necessary permissions

Hmm... That's weird. Everyone should be able to create an issue. Anyway, you already created a pull request #2526. Let me take a look.

takezoe:
@jyoo980 https://github.com/gitbucket/gitbucket/pull/2526 has been merged.

jyoo980:
Cool, I can work on this tomorrow!

jyoo980:
Oops, messed up the commits somehow. I'll open a new PR.

PR #2526, opened 2020-08-26
refactored ActivityService logging strategy

jyoo980:

This is a follow up from the conversation @takezoe and I had in [this conversation](https://github.com/gitbucket/gitbucket/pull/2523#discussion_r476191217)


src/main/scala/gitbucket/core/service/ActivityService.scala
```diff
@@ -97,478 +98,6 @@ trait ActivityService {
...
-        None,
-        currentDate,
-        UUID.randomUUID().toString
-      )
-    )
-  }
-
-  private def cut(value: String, length: Int): String =
-    if (value.length > length) value.substring(0, length) + "..." else value
+  def recordActivity[T <: BaseActivityInfo](info: T): Unit =
```
jyoo980, line 101: This is honestly the main benefit of this work. We no longer need to create a new function each time we want to log something. This function takes care of them all!

src/main/scala/gitbucket/core/service/HandleCommentService.scala
```diff
@@ -29,25 +36,31 @@ trait HandleCommentService {
...
 
-          val (action, actionActivity) = actionOpt
+          actionOpt.collect {
+            case "close" if !issue.closed =>
+              updateClosed(owner, name, issue.issueId, true)
+            case "reopen" if issue.closed =>
+              updateClosed(owner, name, issue.issueId, false)
+          }
+
+          val (action, _) = actionOpt
```
jyoo980, line 46: I could definitely simplify this further, but a combination of not being too familiar with this specific part, and being tired from this refactor is making me put this off until later.

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
```diff
@@ -0,0 +1,95 @@
+package gitbucket.core.model.activity.comment
+
+import java.util.UUID
+
+import gitbucket.core.model.Activity
+import gitbucket.core.model.Profile.currentDate
+import gitbucket.core.model.activity.BaseActivityInfo
+
+sealed trait CommentActivityInfo extends BaseActivityInfo {
```
takezoe, line 9: Even if we have a base trait for specific types of activity information, each sub case classes need to have conductor arguments and we need to implement conversion to `Activity` in each class. Therefore, these base trait would be not much meaningful.

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
```diff
@@ -0,0 +1,95 @@
...
+  def userName: String
+  def repositoryName: String
+  def activityUserName: String
+  def comment: String
+  def maybeIssueId: Option[Int]
+  def maybeCommitId: Option[String]
```
takezoe, line 15: Why do we need to have these `maybeXxxx` methods? It looks no one refer to them.

src/main/scala/gitbucket/core/service/HandleCommentService.scala
takezoe: Okay, I got it. 😅

src/main/scala/gitbucket/core/model/activity/tag/TagActivityInfo.scala
```diff
@@ -0,0 +1,53 @@
+package gitbucket.core.model.activity.tag
```
takezoe, line 1: I feel a little uncomfortable in the package structure of ActivityInfo classes. For example, `TagActivityInfo` and `BranchActivityInfo` should be categorized as `git` separated from `IssueActivityInfo`, etc. Also, `ForkActivityAction` should be categorized as `repository`.

I think simply putting all ActivityInfo classes under `activity` package would be fine because in other types of components like controllers or models also have a flat package structure.

takezoe (commented):
@jyoo980 Overrall, this looks very good to me, yet I left some comments on the structure of ActivityInfo's type hierarchy and packages.

jyoo980:
> @jyoo980 Overall, this looks very good to me, yet I left some comments about the structure of ActivityInfo's type hierarchy and packages.

Thanks so much for all your feedback, @takezoe, I've addressed your comments 🙂

src/main/scala/gitbucket/core/model/activity/tag/TagActivityInfo.scala
jyoo980: That sounds good to me, I can put all the classes under the `activity` package, since this project appears to favour a flat package structure.

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
jyoo980: Right, I understand what you're saying here. 

I think the benefit of this would be future ergonomics and extensibility. An alternative way of doing this would be to
* have each case class extend the `BaseActivityInfo` trait

This could remove the need for our base traits. That being said, I don't think this is the correct approach. I could see a point in the future where we want to have a function that consumes some activity. We could have a function like `logOnlyCommentInfo`. With our first approach of having the base trait, we'd have something like

```Scala
def logOnlyCommentInfo(info: BaseActivityInfo): Unit = info match {
  case 1 => ...
   ...
  case n =>
}
```
of course we could simplify this by using the wildcard operator `_` for pattern matching, but I think that could sometimes be error prone, with our approach of having a sealed trait up top, we could just have
```Scala
def logOnlyCommentInfo(info: CommentActivityInfo): Unit = info match {
  case 1 => ..
  case "last sub type of CommentActivityInfo" => ...
}
```
I would also argue that this is much safer. We get a compile-time error if we fail to exhaustively provide a case for all subtypes of `CommentActivityInfo`.

What are your thoughts?

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
```diff
@@ -0,0 +1,95 @@
...
+
+final case class IssueCommentInfo(
+  userName: String,
+  repositoryName: String,
+  activityUserName: String,
+  comment: String,
+  issueId: Int
+) extends CommentActivityInfo {
+
+  override def maybeIssueId: Option[Int] = Some(issueId)
```
jyoo980, line 30: In reference to [your comment](https://github.com/gitbucket/gitbucket/pull/2526/files/18e3955e5b496d541220b72ae1b02a37d694688d#r478452455), a overrided value is provided here

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
```diff
@@ -0,0 +1,95 @@
...
+  userName: String,
+  repositoryName: String,
+  activityUserName: String,
+  comment: String,
+  commitId: String
+) extends CommentActivityInfo {
+
+  override def maybeIssueId: Option[Int] = None
+
+  override def maybeCommitId: Option[String] = Some(commitId)
```
jyoo980, line 82: In reference to [your comment](https://github.com/gitbucket/gitbucket/pull/2526/files/18e3955e5b496d541220b72ae1b02a37d694688d#r478452455), a overrided value is provided here

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
jyoo980: See https://github.com/gitbucket/gitbucket/pull/2526#discussion_r478525708

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
takezoe: My question was who uses this value? Is this also for future use?

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
takezoe: > I think the benefit of this would be future ergonomics and extensibility.

If we have a plan to have such narrower logging methods, intermediate traits like `CommentActivityInfo` would be beneficial in the near future. I understand the technical benefit of it but I'm wondering whether there is such a possibility in our future.

Anyway, introducing intermediate traits doesn't have a downside while it may be beneficial in the future, so I'm not against introducing intermediate traits.

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
jyoo980: Sorry for not addressing your question properly; this value is necessary for the `CommentActivityInfo` trait to be subtyped by both `IssueCommentInfo` and `CommitCommentInfo`. 

I see that you've also asked a question regarding why the intermediate trait is necessary, and I've provided an explanation [here](https://github.com/gitbucket/gitbucket/pull/2526#discussion_r478523839). In my opinion, this is a small sacrifice in order to have the benefits of exhaustive pattern-matching and compile-time checks

takezoe: I understand your intention of introducing intermediate traits. Regarding these `maybeXxxx` methods, since no one calls these methods currently, so I thought these methods are defined for future use as same as intermediate traits. Is my understanding correct?

My personal preference is adding something when we need them. Therefore, these methods (and intermediate traits) looks redundant for me at the moment, to be honest.

jyoo980: Thanks for the feedback; I guess we have different ideas of how this should look. I'll defer to you since you have many more years of experience with this codebase than I have.

src/main/scala/gitbucket/core/model/activity/comment/CommentActivityInfo.scala
jyoo980: Decided to remove intermediate traits, see https://github.com/gitbucket/gitbucket/pull/2526#discussion_r478752552

takezoe (approved):
Thanks for updating. Looks good to me. 👍

PR #2529, opened 2020-08-27
Add list-public-repositories API

onukura:

## About this PR.

This PR adds list public repositories API.
https://docs.github.com/en/rest/reference/repos#list-public-repositories

but still since param is not suppoted.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/api/ApiRepositoryControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiRepositoryControllerBase.scala
@@ -56,7 +56,7 @@ trait ApiRepositoryControllerBase extends ControllerBase {
 
   /**
    * iv. List all public repositories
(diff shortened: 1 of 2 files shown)


takezoe (approved):
Basically, this looks good to me. However, if GitBucket has too many public repositories, this endpoint can be significantly slow. It would be nice if we can implement pagination as GitHub. It can be a future work.

onukura:
> It would be nice if we can implement pagination as same as GitHub. It can be a future work.

I fully agree with you!

takezoe:
@onukura Could you also update Wiki page? Adding this new endpoint to the list with (since v4.35.0) is fine.
https://github.com/gitbucket/gitbucket/wiki/API-WebHook

onukura:
@takezoe Thank you for merging PR! I added this endpoint to the list in wiki.

PR #2538, opened 2020-08-30
Add update-an-issue-comment API

onukura:

## About this PR

This PR add [API endpoint of "Update an issue comment"](https://developer.github.com/v3/issues/comments/#update-an-issue-comment).

### DIFF
--- a/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
@@ -89,11 +89,14 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
                           issue.isPullRequest
                         )
                       )
+                    case _ => NotFound()
                   }
+                case _ => NotFound()
               }
             } else {
               Unauthorized()
             }
+          case _ => NotFound()
         }
       case _ => NotFound()
     }
--- a/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
@@ -65,9 +65,29 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
   })
 
   /*
-   * v. Edit a comment
-   * https://developer.github.com/v3/issues/comments/#edit-a-comment
+   * v. Update an issue comment
+   * https://developer.github.com/v3/issues/comments/#update-an-issue-comment
    */
+  patch("/api/v3/repos/:owner/:repository/issues/comments/:id")(readableUsersOnly { repository =>
+    (for {
+      commentId <- params("id").toIntOpt
+      issueComment <- getComment(repository.owner, repository.name, commentId.toString)
+      issue <- getIssue(repository.owner, repository.name, issueComment.issueId.toString)
+      body <- extractFromJsonBody[CreateAComment].map(_.body) if !body.isEmpty
+      (issue, id) <- updateCommentByApi(repository, issue, commentId.toString, Some(body))
+      issueComment <- getComment(repository.owner, repository.name, id.toString)
+    } yield {
+      JsonFormat(
+        ApiComment(
+          issueComment,
+          RepositoryName(repository),
+          issue.issueId,
+          ApiUser(context.loginAccount.get),
+          issue.isPullRequest
+        )
+      )
+    }) getOrElse NotFound()
+  })
 
   /*
    * vi. Delete a comment
--- a/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
@@ -69,24 +69,34 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
    * https://developer.github.com/v3/issues/comments/#update-an-issue-comment
    */
   patch("/api/v3/repos/:owner/:repository/issues/comments/:id")(readableUsersOnly { repository =>
-    (for {
-      commentId <- params("id").toIntOpt
-      issueComment <- getComment(repository.owner, repository.name, commentId.toString)
-      issue <- getIssue(repository.owner, repository.name, issueComment.issueId.toString)
-      body <- extractFromJsonBody[CreateAComment].map(_.body) if !body.isEmpty
-      (issue, id) <- updateCommentByApi(repository, issue, commentId.toString, Some(body))
-      issueComment <- getComment(repository.owner, repository.name, id.toString)
-    } yield {
-      JsonFormat(
-        ApiComment(
-          issueComment,
-          RepositoryName(repository),
-          issue.issueId,
-          ApiUser(context.loginAccount.get),
-          issue.isPullRequest
-        )
-      )
-    }) getOrElse NotFound()
+    val commentId = params("id")
+    getComment(repository.owner, repository.name, commentId) match {
+      case Some(issueComment) =>
+        getIssue(repository.owner, repository.name, issueComment.issueId.toString) match {
+          case Some(issue) =>
+            if (isEditable(issue.userName, issue.repositoryName, issue.openedUserName)) {
+              val body = extractFromJsonBody[CreateAComment].map(_.body)
+              updateCommentByApi(repository, issue, issueComment.commentId.toString, body) match {
+                case Some(_) =>
+                  getComment(repository.owner, repository.name, commentId) match {
+                    case Some(issueComment) =>
+                      JsonFormat(
+                        ApiComment(
+                          issueComment,
+                          RepositoryName(repository),
+                          issue.issueId,
+                          ApiUser(context.loginAccount.get),
+                          issue.isPullRequest
+                        )
+                      )
+                  }
+              }
+            } else {
+              Unauthorized()
+            }
+        }
+      case _ => NotFound()
+    }
   })
 
   /*
--- a/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
@@ -66,39 +66,37 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
 
   /*
    * v. Update an issue comment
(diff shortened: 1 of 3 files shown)


src/main/scala/gitbucket/core/plugin/IssueHook.scala
```diff
@@ -13,6 +13,10 @@ trait IssueHook {
     implicit session: Session,
     context: Context
   ): Unit = ()
+  def updatedComment(commentId: Int, content: String, issue: Issue, repository: RepositoryInfo)(
```
jyoo980, line 16: Hi, @onukura 

The work you did looks really good. I'm just wondering what the purpose of this function is. I'm totally new to this project so maybe you can explain this to me.

This method doesn't seem to have an overrided implementation anywhere, what exactly does it do?

onukura: @jyoo980 Hi! actually i am new too ;) So I probably can not answer well for your question. 
This is hook point which will be called when comment update happen. This kind of Hooks are often used by Plugin.
for example gitbucket-notification-plugin use IssueHook

https://github.com/gitbucket/gitbucket-notifications-plugin/blob/b51f194d2b7df2639471a3cc7a91441cb627607c/src/main/scala/gitbucket/notifications/service/NotificationsHook.scala#L57-L79

So I added Hook point for updated comment event as there was not.

@takezoe Is it ok to add `updatedComment` hook point to IssueHook?

jyoo980: Ah so it's actually being overriden by a separate project. I see, thanks so much for the clarification!

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
```diff
@@ -65,9 +65,39 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
...
+  patch("/api/v3/repos/:owner/:repository/issues/comments/:id")(readableUsersOnly { repository =>
+    val commentId = params("id")
+    getComment(repository.owner, repository.name, commentId) match {
+      case Some(issueComment) =>
+        getIssue(repository.owner, repository.name, issueComment.issueId.toString) match {
+          case Some(issue) =>
+            if (isEditable(issue.userName, issue.repositoryName, issue.openedUserName)) {
+              val body = extractFromJsonBody[CreateAComment].map(_.body)
+              updateCommentByApi(repository, issue, issueComment.commentId.toString, body) match {
+                case Some(_) =>
```
jyoo980, line 80: You don't have a case for `None` here, won't this trigger a "non-exhaustive match" compiler error?

jyoo980: Ah, never mind.  You have a catch-all case
```Scala
case _ => NotFound()
```
on line 98

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
```diff
@@ -65,9 +65,39 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
   })
 
   /*
-   * v. Edit a comment
-   * https://developer.github.com/v3/issues/comments/#edit-a-comment
+   * v. Update an issue comment
+   * https://developer.github.com/v3/issues/comments/#update-an-issue-comment
    */
+  patch("/api/v3/repos/:owner/:repository/issues/comments/:id")(readableUsersOnly { repository =>
+    val commentId = params("id")
```
jyoo980, line 72: There's a lot of nesting going on, do you think it would be a better idea to simplify this by doing the `isEditable` check in the `updateCommentByApi` function?

I think the reason you do it here is because you want to preserve the `Unauthorized` error right? Otherwise you only have the possibility of returning `Option[(Issue, Int)]` in `updateCommentByApi`. Maybe you can try using the `Either[A, B]` type here and make `updateCommentByApi` return `Either[ActionResult, Option[String]]`?

Alternatively, maybe you can do the following
```Scala
patch("/api/v3/repos/:owner/:repository/issues/comments/:id")(readableUsersOnly { repository =>
    val commentId = params("id")
    (for {
      issueComment <- getComment(repository.owner, repository.name, commentId)
      issue <- getIssue(repository.owner, repository.name, issueComment.issueId.toString)
    } yield {
      if (isEditable(repository.owner, repository.name, issue.openedUserName)) {
        val maybeBody = extractFromJsonBody[CreateAComment].map(_.body)
        val maybeUpdatedBody = updateCommentWithPermission(repository, issue, issueComment, maybeBody)
        Right(maybeUpdatedBody)
      } else {
        Left(Unauthorized())
      }
    }).map {
        case Right(value) => value.getOrElse(NotFound())
        case Left(value)  => value
      }
      .getOrElse(NotFound())
  })

  private[this] def updateCommentWithPermission(
    repoInfo: RepositoryInfo,
    issue: Issue,
    issueComment: IssueComment,
    maybeBody: Option[String]
  ): Option[String] =
    for {
      _ <- updateCommentByApi(repoInfo, issue, issueComment.commentId.toString, maybeBody)
      issueComment <- getComment(repoInfo.owner, repoInfo.name, issueComment.commentId.toString)
    } yield {
      JsonFormat(
        ApiComment(
          issueComment,
          RepositoryName(repoInfo),
          issue.issueId,
          ApiUser(context.loginAccount.get),
          issue.isPullRequest
        )
      )
    }
```

I'm definitely not a Scala expert, just wanted to get you thoughts on this approach 🙂

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
```diff
@@ -65,9 +65,39 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
   })
 
   /*
-   * v. Edit a comment
-   * https://developer.github.com/v3/issues/comments/#edit-a-comment
+   * v. Update an issue comment
+   * https://developer.github.com/v3/issues/comments/#update-an-issue-comment
```
SIkebe, line 69:
```suggestion
   * https://docs.github.com/en/rest/reference/issues#update-an-issue-comment
```

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
SIkebe: Yeah, there are some compile time warnings...

image

SIkebe (commented):
If the specified comment-id associates with the action other than `comment`, strange things would happen. You may want to filter out those cases.

image

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
onukura: > Yeah, there are some compile time warnings...

yeah. I didn't put `case _ ` because once you find comment, you are sure there is issue. but caused compile warning...
so I fixed this in https://github.com/gitbucket/gitbucket/pull/2538/commits/69f94f435c1fd476a74499c7e0621a167c743df4

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
jyoo980: Any thoughts @onukura ?

onukura: @jyoo980 Thanks for your proposal! Yeah. now my code has so many nests... your way looks simple and nice. I am not scala expert too so I'm not sure which is better... 

@SIkebe @takezoe What do you think about this?

jyoo980: @onukura your way is no better than my way in terms of what it achieves 🙂

The best (and worst) thing about code is that there's so many ways to do the same thing, and it's not always clear which way is the best. The work you've done is _awesome_!!!

takezoe: In this case, the combination of for comprehension and map seems better to reduce nests, but it's just my preference. For sure, as @jyoo980 mentioned, there are many ways to do the same thing in Scala and it's always tough to say which way is better while this flexibility is one beneficial points of Scala.

src/main/scala/gitbucket/core/plugin/IssueHook.scala
takezoe: Adding this hook method is beneficial for plugins. Thanks!

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
takezoe: Also, in this case, using `for` comprehension (or the combination of `flatMap` and `map`) and translate `None` to `NotFound` at last might reduce pattern matching as follows:
```scala
val result = getCommnet(...).flatMap { issueComment =>
  getIssue(...).map { issue =>
    ..
  }
}

result match {
  case Some(response) => response
  case None => NotFound()
}
```

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
```diff
@@ -65,9 +65,42 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
...
    */
+  patch("/api/v3/repos/:owner/:repository/issues/comments/:id")(readableUsersOnly { repository =>
+    val commentId = params("id")
+    getComment(repository.owner, repository.name, commentId) match {
+      case Some(issueComment) =>
+        getIssue(repository.owner, repository.name, issueComment.issueId.toString) match {
+          case Some(issue) =>
+            if (isEditable(issue.userName, issue.repositoryName, issue.openedUserName)) {
```
takezoe, line 77: We need to check whether a comment is editable, not an issue.

onukura: @takezoe  Thx for your review! I modified PR at https://github.com/gitbucket/gitbucket/pull/2538/commits/38bd5d912ce8f793abbbb1ecbef85438400eb7a3.

takezoe (approved):
LGTM 👍

onukura:
@takezoe Thank you for merging this PR.  Wiki updated.

Issue #2539, opened 2020-08-30
Add support for GitHub API v3 delete-comment endpoint


jyoo980:
### Before submitting an issue to GitBucket I have first:

- [x] read the [contribution guidelines](https://github.com/gitbucket/gitbucket/blob/master/.github/CONTRIBUTING.md)
- [x] searched for similar already existing issue
- [x] read the documentation and [wiki](https://github.com/gitbucket/gitbucket/wiki)

This is a `TODO` in `ApiIssueCommentControllerBase.scala` that I want to take on.

onukura:
solved by #2540

PR #2540, opened 2020-08-31
adding support for GitHub v3 delete comment

jyoo980:

### DIFF
--- a/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
@@ -4,6 +4,7 @@ import gitbucket.core.controller.{Context, ControllerBase}
 import gitbucket.core.service._
 import gitbucket.core.util.Implicits._
 import gitbucket.core.util.{ReadableUsersAuthenticator, ReferrerAuthenticator, RepositoryName}
+import org.scalatra.NoContent
 
 trait ApiIssueCommentControllerBase extends ControllerBase {
   self: AccountService
@@ -71,8 +72,18 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
 
   /*
    * vi. Delete a comment
-   * https://developer.github.com/v3/issues/comments/#delete-a-comment
+   * https://docs.github.com/en/rest/reference/issues#delete-an-issue-comment
+   *
    */
+  delete("/api/v3/repos/{owner}/{repo}/issues/comments/:id")(readableUsersOnly { repository =>
+    (for {
+      commentId <- params("id").toIntOpt
+      comment <- getComment(repository.owner, repository.name, commentId.toString)
+      issue <- getIssue(repository.owner, repository.name, comment.issueId.toString)
+    } yield {
+      deleteCommentByApi(repository, comment, issue)
+    }).fold(NotFound())(_ => NoContent())
+  })
 
   private def isEditable(owner: String, repository: String, author: String)(implicit context: Context): Boolean =
     hasDeveloperRole(owner, repository, context.loginAccount) || author == context.loginAccount.get.userName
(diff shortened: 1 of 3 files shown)


src/main/scala/gitbucket/core/service/HandleCommentService.scala
```diff
@@ -134,4 +135,18 @@ trait HandleCommentService {
       }
     }
   }
+
+  def deleteCommentByApi(repoInfo: RepositoryInfo, comment: IssueComment, issue: Issue)(
+    implicit context: Context,
+    s: Session
+  ): Option[(Issue, Int)] = context.loginAccount.flatMap { _ =>
+    val deleteResult = deleteComment(comment.issueId, comment.commentId)
```
jyoo980, line 143: I couldn't find any docs for `deleteComment` here. It returns an `Int` - am I safe to assume that the returned value is the number of records that were deleted in the DB?

If that's the case, then I think making sure it is equal to a non-zero number (or in this case, 1) would be a good idea. Otherwise I don't think there's any need for this `deleteCommentByApi` function to return `Option[(Issue, Int)]`

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
```diff
@@ -71,8 +72,18 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
...
    * vi. Delete a comment
-   * https://developer.github.com/v3/issues/comments/#delete-a-comment
+   * https://docs.github.com/en/rest/reference/issues#delete-an-issue-comment
+   *
    */
+  delete("/api/v3/repos/{owner}/{repo}/issues/comments/:id")(readableUsersOnly { repository =>
+    (for {
+      commentId <- params("id").toIntOpt
+      comment <- getComment(repository.owner, repository.name, commentId.toString)
+      issue <- getIssue(repository.owner, repository.name, comment.issueId.toString)
```
jyoo980, line 82: I'm thinking of doing a permissions check here using `isEditable`, is that a valid approach?

SIkebe (commented):
This implementation allows to delete a comment whatever its action is.
e.g. You can delete an `add_label` comment or a `change_milestone` comment.

You can't delete those comments via Web UI in GitBucket. Does GitHub allow that?

SIkebe:
Closes #2539

takezoe:
> This implementation allows to delete a comment whatever its action is.
> e.g. You can delete an add_label comment or a change_milestone comment.
>
> You can't delete those comments via Web UI in GitBucket. Does GitHub allow that?

Perhaps, GitHub doesn't treat these events as comments? Anyway, we must not allow to delete these event comments in GitBucket even via Web API.

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
takezoe: Yeah, checking whether the user is allowed to delete the comment is necessary. We need to do the same check what we are doing here:
https://github.com/gitbucket/gitbucket/blob/73d05aefadb3e9633818c6dfe40c2a4923f93c6e/src/main/scala/gitbucket/core/controller/IssuesController.scala#L237-L246

src/main/scala/gitbucket/core/controller/api/ApiIssueCommentControllerBase.scala
```diff
@@ -71,8 +72,30 @@ trait ApiIssueCommentControllerBase extends ControllerBase {
...
+   *
    */
+  delete("/api/v3/repos/{owner}/{repo}/issues/comments/:id")(readableUsersOnly { repository =>
+    val maybeDeleteResponse: Option[Either[ActionResult, Option[Int]]] =
+      for {
+        commentId <- params("id").toIntOpt
+        comment <- getComment(repository.owner, repository.name, commentId.toString)
+        issue <- getIssue(repository.owner, repository.name, comment.issueId.toString)
+      } yield {
+        if (isEditable(repository.owner, repository.name, comment.commentedUserName)) {
```
takezoe, line 85: 👍

takezoe (approved):
LGTM 👍

takezoe:
@jyoo980 Sorry, other pull requests I merged before this conflicted with this pull request. Could you resolve the conflicts?

jyoo980:
@takezoe sure, I can take a look this weekend after my classes

takezoe:
Thanks!

takezoe:
@jyoo980 Could you also add this endpoint to the wiki page?
https://github.com/gitbucket/gitbucket/wiki/API-WebHook

jyoo980:
@takezoe I've added the endpoint to the wiki page. I think we can close this.

PR #2546, opened 2020-09-08
Add list-milestones and get-a-milestone API

onukura:

## About this PR

This PR add api endpoints for milestones below.

- [List milestones](https://docs.github.com/en/rest/reference/issues#list-milestones)
- [Get a milestone](https://docs.github.com/en/rest/reference/issues#get-a-milestone)
- [Delete a milestone](https://docs.github.com/en/rest/reference/issues#delete-a-milestone)

### DIFF
--- a/src/main/scala/gitbucket/core/api/ApiMilestone.scala
+++ b/src/main/scala/gitbucket/core/api/ApiMilestone.scala
@@ -1,0 +1,52 @@
+package gitbucket.core.api
+
+import gitbucket.core.model.{Milestone, Repository}
+import gitbucket.core.util.RepositoryName
+import java.util.Date
+
+/**
+ * https://docs.github.com/en/rest/reference/issues#milestones
+ */
+case class ApiMilestone(
+  url: ApiPath,
+//  html_url: ApiPath,
+//  label_url: ApiPath,
+  id: Int,
+  number: Int,
+  state: String,
+  title: String,
+  description: String,
+  creator: ApiUser,
+  open_issues: Int,
+  closed_issues: Int,
+//  created_at: Option[Date],
+//  updated_at: Option[Date],
+  closed_at: Option[Date],
+  due_on: Option[Date]
+)
+
+object ApiMilestone {
+  def apply(
+    repository: Repository,
+    milestone: Milestone,
+    user: ApiUser,
+    milestone_number: Int,
+    open_issue_count: Int = 0,
+    closed_issue_count: Int = 0
+  ): ApiMilestone =
+    ApiMilestone(
+      url = ApiPath(s"/api/v3/repos/${RepositoryName(repository).fullName}/milestones/${milestone_number}"),
+//      html_url = ApiPath(s"/${RepositoryName(repository).fullName}/milestones/${milestone.title}"),
+//      label_url = ApiPath(s"/api/v3/repos/${RepositoryName(repository).fullName}/milestones/${milestone_number}/labels"),
+      id = milestone.milestoneId,
+      number = milestone_number,
+      state = if (milestone.closedDate.isDefined) "closed" else "open",
+      title = milestone.title,
+      description = milestone.description.getOrElse(""),
+      creator = user,
+      open_issues = open_issue_count,
+      closed_issues = closed_issue_count,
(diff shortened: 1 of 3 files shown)


src/main/scala/gitbucket/core/api/ApiMilestone.scala
```diff
@@ -0,0 +1,52 @@
...
+import gitbucket.core.model.{Milestone, Repository}
+import gitbucket.core.util.RepositoryName
+import java.util.Date
+
+/**
+ * https://docs.github.com/en/rest/reference/issues#milestones
+ */
+case class ApiMilestone(
+  url: ApiPath,
+//  html_url: ApiPath,
```
takezoe, line 12: Why can't we provide `html_url`?

onukura: @takezoe I thought gitbucket does not have url like `/:owner/:repository/milestone/:id`. So I didn't provide...
Maybe we can provide `/owner/:repository/issues?milestone={name}&state=open`  ...?
.

takezoe: For now, `/owner/:repository/issues?milestone={name}&state=open` is the only possible URL for milestones. However, we may need to have anothero URL which shows the status of the milestone.

onukura: @takezoe fixed PR.  Now use `/owner/:repository/issues?milestone={name}&state=open` for `html_url`.
> we may need to have anothero URL which shows the status of the milestone.
Yeah it would be nice if GitBucket has each milestone page like GitHub that would be another issue.

src/main/scala/gitbucket/core/controller/api/ApiIssueMilestoneControllerBase.scala
```diff
@@ -0,0 +1,82 @@
...
+        ApiMilestone(
+          repository.repository,
+          milestoneWithIssue._1,
+          ApiUser(context.loginAccount.get),
+          i + 1,
+          milestoneWithIssue._2,
+          milestoneWithIssue._3
+        )
+      }).reverse
+    state match {
```
onukura, line 36: As there is no index correspond to milestone `number` of each repository, I use zipIndex to put number after sorting by `milestoneId` in repository. But actually if some milestones deleted, `number` will be different...? So means this way is not immutable...
Maybe number column is required?? or use `milestoneId` as `number`??

takezoe: I'm not sure the exact meaming of `number` in GitHub API. However, we use `issueId` as `number` in Issue/Pull request APIs. Therefore, using `milestoneId` as `number` makes sense.
https://github.com/gitbucket/gitbucket/blob/f6a5def6386da6d82ffec3af25cdc7366fd456f5/src/main/scala/gitbucket/core/api/ApiIssue.scala#L48-L49

But why `id = 0` in Issue/Pull request APIs...? 🤔 
https://github.com/gitbucket/gitbucket/blob/f6a5def6386da6d82ffec3af25cdc7366fd456f5/src/main/scala/gitbucket/core/api/ApiIssue.scala#L22

onukura: @takezoe Ok! So I will fix PR to use `milestoneId` as `number`.
About `id = 0`, I don't know neither. It is not in use. so we probably can remove...

takezoe: Yeah... Regarding Issue/Pull request API, I think `id` field is necessary, but we need to look into the reason for the dummy value, this is another topic though.

onukura: @takezoe I fixed PR (added delete endpoint)

takezoe: Thanks!

onukura:
wiki updated

Issue #273, opened 2014-02-10
Code syntax highlighting


cuekoo:
Hi! It would be great to add syntax highlighting support for languages like `matlab`. And I am also wondering is it possible to change the highlighting style (for example change to `github` style) ?

PR #2555, opened 2020-09-16
Add syntax highlighter setting per user

onukura:

## About this PR

This PR add syntax highlighter theme setting per user.
Themes are from https://github.com/jmblog/color-themes-for-google-code-prettify.

pr_screenshot_list


src/main/resources/update/gitbucket-core_4.35.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<changeSet>
+  <!--================================================================================================-->
+  <!-- ACCOUNT_HIGHLIGHTER -->
+  <!--================================================================================================-->
+  <createTable tableName="ACCOUNT_HIGHLIGHTER">
```
takezoe, line 6: If we will have per-user preferences in GitBucket, it might be better to define a table like `ACCOUNT_PREFERENCES` or `USER_PREFERENCES` so that we can add columns for another per-user preference items to this table later.

src/main/scala/gitbucket/core/service/AccountHighlighterService.scala
```diff
@@ -0,0 +1,34 @@
+package gitbucket.core.service
+
+import gitbucket.core.model.AccountHighlighter
+import gitbucket.core.model.Profile._
+import gitbucket.core.model.Profile.profile.blockingApi._
+
+trait AccountHighlighterService {
```
takezoe, line 7: I think handling per-user preferences can be parts of `AccountService` because it will highly possibly have further per-user preferences in the future and in that case, adding new service trait per preference item is annoying.

src/main/resources/update/gitbucket-core_4.35.xml
takezoe, line 6: If we will have per-user preferences mechanism in GitBucket, it might be better to define a table like `ACCOUNT_PREFERENCES` or `USER_PREFERENCES` so that we can add columns for another per-user preference items to this table later.

By the way, GitBucket currently uses the browser's localStorage to store a line wrapping setting in the online editor. We may be able to use the same method for highlighter setting by putting a pulldown on the top of the source code viewer without adding a table.

I can't conclude which is a better solution immediately. I just want to let you know we have another option to store per-user settings for now.

src/main/scala/gitbucket/core/service/AccountHighlighterService.scala
onukura: @takezoe 
> adding new service trait per preference item is annoying.
Yeah. that make sense. so I will put them into `AccountService`!

src/main/resources/update/gitbucket-core_4.35.xml
onukura: @takezoe 
> GitBucket currently uses the browser's localStorage to store a line wrapping setting in the online editor. We may be able to use the same method for highlighter setting by putting a pulldown on the top of the source code viewer without adding a table.

I didn't know this system... 😅 that could be choice too.

hmm.. IMHO user preferences table might be useful considering that we probably want to add more user preferences and if it does not belongs to a browser, setting reproduce regardless user's browser and user's device.

But As I said before I didn't know browser storage, so I'm really not sure which is better...

takezoe: I think we need to sort out what settings should belong to browser or user. Line wrapping setting may also need to be migrated to user preferences if we will introduce the per-user preferences table.

For now, could you update the table schema in case of storing further per-user preferences in the future? Then, I will test it and consider whether we should migrate other preferences currently use localStorage to the per-user preferences table.

onukura: @takezoe I updated table schema and services.  Thank you for your consideration.

takezoe: Thanks! I will try this PR.

takezoe:
I tried this pull request and also checked GitHub's behavior:

- Online editor settings (soft wrap/no wrap, etc) seem to be not persisted
- Pull request presentation settings (split/unified, etc) seem to be persisted per account

I agree with persisting the editor syntax highlighting setting as an account setting. In addition, I would like to persist pull request settings per account if the user is signed-in in the future.

onukura:
@takezoe Thanks for your consideration and merge PR!!

> I would like to persist pull request settings per account if the user is signed-in in the future.

Sounds good!

onukura:
related to #273

PR #2557, opened 2020-09-23
Add update a pull request API

onukura:

## About this PR

This PR add [Update a pull request](https://docs.github.com/en/rest/reference/pulls#update-a-pull-request) API.

### DIFF
--- a/src/main/scala/gitbucket/core/service/PullRequestService.scala
+++ b/src/main/scala/gitbucket/core/service/PullRequestService.scala
@@ -320,16 +320,36 @@ trait PullRequestService {
         if (Repositories.filter(_.byRepository(pr.userName, pr.repositoryName)).exists.run) {
           // Update base branch
           base.foreach { _base =>
-            Using.resource(Git.open(getRepositoryDir(repository.owner, repository.name))) { git =>
-              getBranches(git, repository.repository.defaultBranch, origin = true)
-                .find(_.name == _base)
-                .foreach(br => updateBaseBranch(repository.owner, repository.name, issueId, br.name, br.commitId))
+            if (pr.branch != _base) {
+              Using.resource(Git.open(getRepositoryDir(repository.owner, repository.name))) { git =>
+                getBranches(git, repository.repository.defaultBranch, origin = true)
+                  .find(_.name == _base)
+                  .foreach(br => updateBaseBranch(repository.owner, repository.name, issueId, br.name, br.commitId))
+              }
+              createComment(
+                repository.owner,
+                repository.name,
+                loginAccount.userName,
+                issue.issueId,
+                pr.branch + "\r\n" + _base,
+                "change_base_branch"
+              )
             }
           }
           // Update title and content
-          title.foreach(
-            _title => updateIssue(repository.owner, repository.name, issueId, _title, body)
-          )
+          title.foreach { _title =>
+            updateIssue(repository.owner, repository.name, issueId, _title, body)
+            if (issue.title != _title) {
+              createComment(
+                repository.owner,
+                repository.name,
+                loginAccount.userName,
+                issue.issueId,
+                issue.title + "\r\n" + _title,
+                "change_title"
+              )
+            }
+          }
           // Update state
           val action = (state, issue.closed) match {
             case (Some("open"), true) =>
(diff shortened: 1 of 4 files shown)


src/main/scala/gitbucket/core/service/PullRequestService.scala
```diff
@@ -293,6 +302,66 @@ trait PullRequestService {
...
+    c: JsonFormat.Context
+  ): Unit = {
+    getPullRequest(repository.owner, repository.name, issueId) match {
+      case Some(x) =>
+        if (Repositories.filter(_.byRepository(x._2.userName, x._2.repositoryName)).exists.run) {
+          // Update base branch
+          base match {
+            case Some(_base) =>
+              Using.resource(Git.open(getRepositoryDir(repository.owner, repository.name))) { git =>
+                getBranches(git, repository.repository.defaultBranch, origin = true).find(_.name == _base) match {
```
takezoe, line 325: In this case, using `foreach` instead of `match` would be better so that we can get rid of `case _ =>`.

src/main/scala/gitbucket/core/service/PullRequestService.scala
```diff
@@ -293,6 +302,66 @@ trait PullRequestService {
...
+    base: Option[String]
+  )(
+    implicit s: Session,
+    c: JsonFormat.Context
+  ): Unit = {
+    getPullRequest(repository.owner, repository.name, issueId) match {
+      case Some(x) =>
+        if (Repositories.filter(_.byRepository(x._2.userName, x._2.repositoryName)).exists.run) {
+          // Update base branch
+          base match {
```
takezoe, line 322: In this case, using `foreach` instead of `match` would be better so that we can get rid of `case _ =>`.

src/main/scala/gitbucket/core/service/PullRequestService.scala
```diff
@@ -293,6 +302,66 @@ trait PullRequestService {
...
+                  case Some(br) =>
+                    updateBaseBranch(repository.owner, repository.name, issueId, br.name, br.commitId)
+                  case _ =>
+                }
+              }
+            case _ =>
+          }
+
+          // update title and content
+          title match {
```
takezoe, line 335: Ditto

src/main/scala/gitbucket/core/service/PullRequestService.scala
```diff
@@ -293,6 +302,66 @@ trait PullRequestService {
...
+    settings: SystemSettings,
+    title: Option[String],
+    body: Option[String],
+    state: Option[String],
+    base: Option[String]
+  )(
+    implicit s: Session,
+    c: JsonFormat.Context
+  ): Unit = {
+    getPullRequest(repository.owner, repository.name, issueId) match {
```
takezoe, line 318: In this case, using `foreach` instead of `match` would be better so that we can get rid of `case _ =>`.

src/main/scala/gitbucket/core/service/PullRequestService.scala
```diff
@@ -293,6 +302,66 @@ trait PullRequestService {
...
+      case Some(x) =>
```
takezoe, line 319: Ditto

src/main/scala/gitbucket/core/service/PullRequestService.scala
onukura: @takezoe Thanks for your review! I fixed PR!
BTW I think we should add function to let message when repository updated by api. Currently api can change (PR, Issue, Issue comment, etc. ) silently... that causes lack of transparency.

takezoe: > BTW I think we should add function to let message when repository updated by api.

A kind of auditing? Do you have any idea where or how such messages will be shown?

onukura: When user change tag, we can see history on issue like below.

image

IMHO It would be more transparent if this kind of history show up even when some thing got changed by api.
But it is another issue.

onukura: I noticed that I just can use [createComment](https://github.com/gitbucket/gitbucket/blob/8dbcbb54851feffca390957990baffc43697b08d/src/main/scala/gitbucket/core/service/IssuesService.scala#L516) to leave history.
So I updated my PR to leave comment. https://github.com/gitbucket/gitbucket/pull/2557/commits/b7506459a5b3c02d0c0d84f460787996c581b582

takezoe: Great thanks!

onukura:
wiki updated

PR #2574, opened 2020-10-07
Add Get-branch-protection, Delete-branch-protection, Get-all-status-check-contexts API

onukura:

## About this PR

This PR adds Web API endpoints below

- [Get branch protection](https://docs.github.com/en/rest/reference/repos#get-branch-protection)
- [Delete branch protection](https://docs.github.com/en/rest/reference/repos#delete-branch-protection)
- [Get all status check contexts](https://docs.github.com/en/rest/reference/repos#get-all-status-check-contexts)
- [Get status checks protection](https://docs.github.com/en/rest/reference/repos#get-status-checks-protection)

### DIFF
--- a/src/main/scala/gitbucket/core/api/ApiBranchProtection.scala
+++ b/src/main/scala/gitbucket/core/api/ApiBranchProtection.scala
@@ -4,7 +4,10 @@ import gitbucket.core.service.ProtectedBranchService
 import org.json4s._
 
 /** https://developer.github.com/v3/repos/#enabling-and-disabling-branch-protection */
-case class ApiBranchProtection(enabled: Boolean, required_status_checks: Option[ApiBranchProtection.Status]) {
+case class ApiBranchProtection(
+  enabled: Boolean,
+  required_status_checks: Option[ApiBranchProtection.Status]
+) {
   def status: ApiBranchProtection.Status = required_status_checks.getOrElse(ApiBranchProtection.statusNone)
 }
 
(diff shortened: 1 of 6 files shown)


onukura:
Ops. I should have updated test case...

takezoe:
@onukura There is conflict with #2575. Could you resolve it?

onukura:
@takezoe Conflict fixed!

onukura:
I added [Get status checks protection](https://docs.github.com/en/rest/reference/repos#get-status-checks-protection) endpoint.

takezoe:
Thanks!

takezoe:
Hmm... This pull request seems to broke the branch protection setting form. Status checking setting is not persistent in the current master.

takezoe:
`ApiPath` cannot be used for input case classes. In this case, `ApiBranchProtection` is used for both input and output and `ApiPath` is used in `Status` without wrapping by `Option`. Therefore, it seems to fail to deserialize JSON to `Status`. We have two options to solve the issue here:

- Use different case class for input/output instead of `ApiBranchProtection`
- Wrap `ApiPath` by `Option` and give `Some` value only for output

Also, JSON serialization/deserialization test would be necessary to avoid this kind of implicit corruption.

takezoe:
Created a pull request: https://github.com/gitbucket/gitbucket/pull/2600

onukura:
@takezoe Thank you for your work to fix my PR! I should have noticed before...

PR #2578, opened 2020-10-10
Integration test for Web API

takezoe:

### DIFF
--- a/src/test/scala/gitbucket/core/api/ApiIntegrationTest.scala
+++ b/src/test/scala/gitbucket/core/api/ApiIntegrationTest.scala
@@ -1,0 +1,20 @@
+package gitbucket.core.api
+
+import gitbucket.core.TestingGitBucketServer
+import gitbucket.core.util.HttpClientUtil
+import org.scalatest.funsuite.AnyFunSuite
+
+import scala.util.Using
+
+class ApiIntegrationTest extends AnyFunSuite {
+
+  test("API integration test") {
+    Using.resource(new TestingGitBucketServer(8080)) { server =>
+      server.start()
+
+      HttpClientUtil.withHttpClient(None) { httpClient =>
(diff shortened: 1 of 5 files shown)


takezoe:
@onukura This can be a base of integration tests for Web API. What do you feel about this?

src/test/scala/gitbucket/core/api/ApiIntegrationTest.scala
```diff
@@ -0,0 +1,73 @@
...
+        assert(repository.getForks == 0)
+        assert(repository.getForksCount == 0)
+        assert(repository.isPrivate == false)
+        assert(repository.getOwner.getLogin == "root")
+        assert(repository.hasIssues == true)
+        assert(repository.getUrl.toString == s"http://localhost:${server.port}/api/v3/repos/root/test")
+        assert(repository.getHttpTransportUrl == s"http://localhost:${server.port}/git/root/test.git")
+        assert(repository.getHtmlUrl.toString == s"http://localhost:${server.port}/root/test")
+      }
+    }
```
onukura, line 70: @takezoe At first thank you for your work about API test basement! 
I guess we will need add this kind of part for each endpoint, right?? It might be a lot of work, but it is also what we should do for reliability of API. I agree with your idea!

onukura:
And I really need to know about this framework because I have no idea if endpoint is more complicated one like PR and StatusChecks...

takezoe:
I don't yet have an idea for the complicated cases. I would like to try such cases in this pull request to find a way.

takezoe:
Added test cases for pull request status as examples of complex test cases.

By the way, this integration test requires to run `./sbt package` in advance to setup web application resources for the embedded Jetty. However, I think it would be nice if we can resolve it without running sbt.

Issue #2146, opened 2018-09-19
Collaborators can delete Issue Comments


PaulAdams-Conti:
## Issue
**Impacted version**: 4.28.0

**Deployment mode**: Tomcat 7

**Problem description**:
Collaborators can delete any Issue Comment (maybe related to #1462). IMO, only repo admins should be able to delete comments otherwise issue traceability is compromised.

Also unlike in GitHub, there is no record that anyone has deleted a comment, it just disappears.

It would be good if there was some kind of history for the comments so that if necessary, deleted comments can be retrieved in some way.

So my prioritised feature request is as follows:

1. Only repo admins can delete comments.
2. If a comment is deleted, a message should be shown inline with the comments that the comment was deleted and by whom (like in GitHub).
3. Some retrieval possibility or history log for deleted comments at least for repo admins.

PR #2588, opened 2020-10-27
Improve issue comment management

onukura:

## About this PR

This PR is related to #2146 .
By this PR, issue comment behavior will be like below

- Only repo admins and comment author can delete comments. (developer, guest are not allowed)
- If a comment is deleted, a message will be shown.

"restore comment" function is not so easy to implement, so this PR does not handle this.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/IssuesController.scala
+++ b/src/main/scala/gitbucket/core/controller/IssuesController.scala
@@ -240,7 +240,7 @@ trait IssuesControllerBase extends ControllerBase {
       case (owner, name) =>
         getComment(owner, name, params("id")).map { comment =>
           if (isEditableContent(owner, name, comment.commentedUserName)) {
-            Ok(deleteComment(comment.issueId, comment.commentId))
+            Ok(deleteComment(repository.owner, repository.name, comment.issueId, comment.commentId))
           } else Unauthorized()
         } getOrElse NotFound()
     }
--- a/src/main/scala/gitbucket/core/controller/IssuesController.scala
+++ b/src/main/scala/gitbucket/core/controller/IssuesController.scala
@@ -239,7 +239,7 @@ trait IssuesControllerBase extends ControllerBase {
     defining(repository.owner, repository.name) {
       case (owner, name) =>
         getComment(owner, name, params("id")).map { comment =>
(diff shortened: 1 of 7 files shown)


src/main/scala/gitbucket/core/controller/IssuesController.scala
```diff
@@ -239,7 +240,7 @@ trait IssuesControllerBase extends ControllerBase {
       case (owner, name) =>
         getComment(owner, name, params("id")).map { comment =>
           if (isEditableContent(owner, name, comment.commentedUserName)) {
```
takezoe, line 242: Therefore, `isEditableContent` might not be not enough?

onukura: @takezoe You mean we need something like `isDeletableContent`?? 

Acutually this ajaxPost endpoint is not limited as on UI. So means.,,
On ajaxPost Endpoint: developer role can delete comment.
On UI: developer role can not see delete button.
-> finally developer role can not delete comment.

takezoe: > this ajaxPost endpoint is not limited as on UI.

I see... So, this pull request is just willing to hide a delete button on UI. I think it's better to make those consistent. Let me check who calls this ajax endpoint before concluding this topic.

takezoe: Looks like this ajax endpoint is called from only the delete button.

https://github.com/gitbucket/gitbucket/blob/08ed3c4171373d8f6bad2c8ddbb7b393c6fc309e/src/main/twirl/gitbucket/core/issues/commentlist.scala.html#L318-L328

This means we can restrict this endpoint for only who can see the delete button?

onukura: @takezoe I introduced `isDeletableComment` function at https://github.com/gitbucket/gitbucket/pull/2588/commits/be640d9539dcd964fa6c51e46dae62e1b12fc82c. Now UI and delete ajaPost are consistent.

takezoe: Thanks!!

Issue #1429, opened 2017-01-17
Request: Hide navbar if anonymous access is active


zy4:
The option description states that on "Deny" : 

> " Users must authenticate before viewing any information."

I think that there is no reason for the navbar and the search form to show up in this case.

takezoe:
Fixed by #2589

PR #2589, opened 2020-10-27
Hide header content on signin and register page

### LINKED ISSUE #2058: Hide search bar on login screen
## Issue
**Impacted version**: 4.25.0

**Deployment mode**: standalone

**Problem description**:
A further security improvement that could be done to the login screen is hiding the search bar, thus preventing potentially malicious input from being fed to the app anonymously.

capture

onukura:

## About this PR

This PR resolve #2058 .

User will be not able to see header contents (search bar, menu made by plugin, ...) on sign-in page and register page.

image


src/main/twirl/gitbucket/core/main.scala.html
```diff
@@ -1,4 +1,4 @@
-@(title: String, repository: Option[gitbucket.core.service.RepositoryService.RepositoryInfo] = None)(body: Html)(implicit context: gitbucket.core.controller.Context)
+@(title: String, repository: Option[gitbucket.core.service.RepositoryService.RepositoryInfo] = None, showHeaderContents: Boolean = true)(body: Html)(implicit context: gitbucket.core.controller.Context)
```
onukura, line 1: At first I tried to hide header content (search bar, plugins, ...) by checking user is logged-in and allowAnonymousAccess setting but that causes a case that anonymous user can not see header content on every pages even if allowAnonymousAccess is true.
So I introduced `showHeaderContents` flag to `main.scala.html`.

takezoe:
I guess the original motivation of #2058 is that hiding these elements when GitBucket is configured to prevent anonymous access. Therefore, it might be better to hide these elements in the all pages not only in the sign-in page and the registration page when GitBucket is configured as the private mode and user is not logged-in.

What do you think?

aadrian:
> I guess the original motivation of #2058 is that hiding these elements when GitBucket is configured to prevent anonymous access. 

@takezoe  since the main use case of GitBucket is as a private installation (GitHub Entrprise alternative) - that issue is mainly about security.

onukura:
> it might be better to hide these elements in the all pages not only in the sign-in page and the registration page when GitBucket is configured as the private mode and user is not logged-in.

@takezoe in that case, even if GitBucket is public mode, anonymous user can not see snippets for example whereas repository is still visible.

takezoe:
>  anonymous user can not see snippets

That depends on the plugin's specification. Rather, the visibility of the header menu should be able to be controlled on pluigin side.

aadrian:
>  Do you mean that the motivation of #2058 is different?

@takezoe No. I just wanted to emphasize that indeed: the main reason is security (not esthetics)

takezoe:
@aadrian 

> the main reason is security

That is my understanding, so I proposed that hiding these elements at all pages when users are not logged-in in the private mode. Is there any downside in this idea?

aadrian:
> Is there any downside in this idea?

@takezoe I'm not aware that for **private mode** this could bring any downside.

onukura:
@takezoe @aadrian I updated behavior. 

Now hide header elements in the all pages including sign-in page and the registration page when GitBucket is configured as the private mode and user is not logged-in.

onukura:
also related to #1429

takezoe (approved):

Issue #2152, opened 2018-09-30
Freeze repository (make repository read-only)


takezoe:

controlade:
Under Settings, danger zone:

![image](https://user-images.githubusercontent.com/7881851/56814213-5105c000-6804-11e9-998a-05173b3d86c7.png)

sidcool1234:
> Under Settings, danger zone:
> 
> ![image](https://user-images.githubusercontent.com/7881851/56814213-5105c000-6804-11e9-998a-05173b3d86c7.png)

Can the issue be closed?

takezoe:
This is not implemented yet.

sidcool1234:
> This is not implemented yet.

Thanks

PR #2595, opened 2020-11-11
Add function to archive repository

onukura:

## About this PR

This PR add feature to archive repository (read-only).  This implements #2152. 

### Intended Behavior

If repo is archived, 
- user can read repo. only except public/private setting.
- buttons or input area related to repo. update action will be invisible.
- API end points related to repo. update action will be unavailable.

### Technical changes

- Add `ARCHIVED` column to `REPOSITORY` table. => use like `repositpry.repository.isArchived`
- Introduce `UnarchivedAuthenticator` to reject request on server side if repo. is archived.
- Leave activity log when repo archive statement is changed.

### Screenshot

pr_screenshot_list


src/main/scala/gitbucket/core/util/Authenticator.scala
```diff
@@ -171,3 +171,26 @@ trait GroupManagerAuthenticator { self: ControllerBase with AccountService =>
     }
   }
 }
+
+/**
+ * Allow action on only unarchived repository
+ */
+trait UnarchivedAuthenticator { self: ControllerBase with RepositoryService =>
```
takezoe, line 178: We will have to ensure applying this authenticator to all possible write actions even in plugins, but ensuring it in all plugins is actually difficult or impossible.

One idea I came up with is that checking whether a repository is archived in `writableUsersOnly` too. This is redundant and don't cover all cases, but will automatically apply the minimum protection for the existing plugins. Or adding an option like `allowUnarchivedRepository` to all authenticator so that all actions currently use authenticator will need to be updated to specify that option explicitly.

Anyway, we would need to consider more carefully about how to apply the protection for archived repositories.

src/main/scala/gitbucket/core/service/RepositoryService.scala
```diff
@@ -301,6 +305,39 @@ trait RepositoryService {
...
+    Repositories
+      .filter(_.byRepository(owner, repository))
+      .map { t =>
+        t.isArchived
+      }
+      .update(false)
+  }
+
+  def isArchivedRepository(owner: String, repository: String)(implicit s: Session): Option[Boolean] =
+    getRepository(owner, repository) match {
```
takezoe, line 331: This can be as follows:
```scala
getRepository(owner, repository).map { repository =>
  repository.repository.isArchived
}
```

src/main/scala/gitbucket/core/util/Authenticator.scala
onukura: @takezoe Thanks for you review! I agree that applying `UnarchivedAuthenticator` to all write action is difficult and impossible.

> checking whether a repository is archived in writableUsersOnly

That might be simple and good solution, but actually checking in `writableUserOnly` is not enough. Sometime we want to restrict even `readableUserOnly` user action if repository is archived such as "create new issue". 

> adding an option like allowUnarchivedRepository to all authenticator so that all actions currently use authenticator will need to be updated to specify that option explicitly.

It is more flexible than checking in `writableUsersOnly`, but as you said need update all actions, and even plugins... 

hmm...

src/main/resources/update/gitbucket-core_4.35.xml
```diff
@@ -18,4 +18,8 @@
   </createTable>
   <addPrimaryKey constraintName="IDX_ACCOUNT_PREFERENCE_PK" tableName="ACCOUNT_PREFERENCE" columnNames="USER_NAME"/>
   <addForeignKeyConstraint constraintName="IDX_ACCOUNT_PREFERENCE_FK0" baseTableName="ACCOUNT_PREFERENCE" baseColumnNames="USER_NAME" referencedTableName="ACCOUNT" referencedColumnNames="USER_NAME" onDelete="CASCADE" onUpdate="CASCADE"/>
+
+  <addColumn tableName="REPOSITORY">
+    <column name="ARCHIVED" type="boolean" nullable="false"/>
```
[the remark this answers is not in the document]
onukura: @takezoe fixed in https://github.com/gitbucket/gitbucket/pull/2595/commits/efd49cf428a330792696634a2ace93738595c7e6

src/main/scala/gitbucket/core/service/RepositoryService.scala
onukura: @takezoe fixed in https://github.com/gitbucket/gitbucket/pull/2595/commits/afe6e74b4cb34af6f56907c84e27883886a506bb

src/main/scala/gitbucket/core/service/RepositoryService.scala
```diff
@@ -844,4 +881,20 @@ object RepositoryService {
...
+    override def preReceive(
+      owner: String,
+      repository: String,
+      receivePack: ReceivePack,
+      command: ReceiveCommand,
+      pusher: String
+    )(implicit session: Session): Option[String] = {
+      isArchivedRepository(owner, repository) match {
+        case Some(true)  => Some("This repository was archived so it is read-only.")
+        case Some(false) => None
```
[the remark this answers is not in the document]
onukura: @takezoe fixed in https://github.com/gitbucket/gitbucket/pull/2595/commits/76dab825ed80cb9a39e9df0348a5e153045b7616

src/main/scala/gitbucket/core/util/Authenticator.scala
onukura: > checking whether a repository is archived in `writableUsersOnly` too

@takezoe I misunderstood what you meant about this. You mean check if repository is archived in `writableUsersOnly` and use `UnarchivedAuthenticator` for other cases such as "create new issue" (`readableUserOnly`) if restriction required.

In that case this idea sound better because it "will automatically apply the minimum protection for the existing plugins" as you said.

takezoe: Yes, that is what I meant. But that is just one option. We would need to see more closely and consider possible options carefully to figure out what is the best way for this.

onukura: > We would need to see more closely and consider possible options carefully to figure out what is the best way for this.

@takezoe Yeah. I agree with that.

BTW I updated PR https://github.com/gitbucket/gitbucket/pull/2595/commits/6653dbb8d0e3852b10e2cc739970e959ca870296, https://github.com/gitbucket/gitbucket/pull/2595/commits/89de8fdc9394bb8a3e644cde7164f92206c292a7 because checking in `writableUsersOnly` too sounds better than before at least.

PR #2660, opened 2021-01-27
As is, the unsuspend API cannot unsuspend since it's not searching su…

deadserious:
…spended accounts. This should allow the api to function without getting a 404.


deadserious:
I found that the DELETE supended api will always return 404 for valid people because it was not allowing the search to happen across removed accounts.  Please merge this to fix this bug.

takezoe:
Thanks!

Issue #1694, opened 2017-09-02
Migration from H2 to MariaDB: SQLSyntaxErrorException: (conn:72) Query was empty


uli-heller:
## Issue
**Impacted version**: 4.16.0 (and probably others before as well)

**Deployment mode**: standalone app with an http frontend (httpd)

**Problem description**:

When upgrading to 4.16.0, I decided to migrate from H2 to MariaDB.
The migration was quite simple. I followed the instructions on the wiki page.
Unfortunately, I was not able to import the exported data via the admin
gui interface. I got the error "SQLSyntaxErrorException: (conn:72) Query was empty"
within the log file and on the browser.

Importing the data via command line works perfecty, i.e.

* `mysql -u{user} -p{password} gitbucket </tmp/gitbucket-export-143901791709498677.sql`

takezoe:
@uli-heller Hmm... Could you find a failed query?

solsticesurfer:
I experienced this same error today. The cause was an empty line in the exported SQL.

- Starting from the built-in H2 instance with a fresh install, I added a project and a couple of test issues. No 3rd party plugins were installed.
- Went to the admin settings and exported ALL tables
- The resulting export had an empty line at the end of the SQL file.
- Restart Gitbucket with it running against MySql
- Import the SQL file and the import fails with "Query was empty" error message
- Edit SQL file, remove empty line at the end.
- Re-run import and it works without issue.

Hope this helps other people that may have run up against this issue.

solsticesurfer:
Tried to submit a PR to resolve this but it fails with `(Compile / scalafmtCheck) 1 files must be formatted`. Code appears formatted correctly, it was just a simple if/then check. Haven't found the formatting rules for this project so ¯\_ (ツ)_/¯

takezoe:
@solsticesurfer Sorry for late response and thanks for sending a pull request! I was super busy at work last one week. 😭  Maybe we should identify it and eliminate the cause of empty SQL, but just skipping empty SQL is easier and effective to avoid unpredictable troubles like this. Could you just run `sbt scalafmt` to fix source code format?

takezoe:
Closed as #2664 has been merged.

PR #2664, opened 2021-01-29
Prevent execution of empty SQL strings during database import

solsticesurfer:
This is to address issue #1694 which can result in the execution of an empty SQL string, preventing the import database process from completing successfully.

### DIFF
--- a/src/main/scala/gitbucket/core/util/JDBCUtil.scala
+++ b/src/main/scala/gitbucket/core/util/JDBCUtil.scala
@@ -84,7 +84,7 @@ object JDBCUtil {
               if (c == ';' && !stringLiteral) {
                 val sql = new String(out.toByteArray, "UTF-8")
                 if (sql != null && !sql.isEmpty()) {
-                  conn.update(sql.trim)                 
+                  conn.update(sql.trim)
                 }
                 out = new ByteArrayOutputStream()
               } else {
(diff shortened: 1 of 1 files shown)


takezoe:
@solsticesurfer Hi! Could you `sbt scalafmt` to fix source code format?

src/main/scala/gitbucket/core/util/JDBCUtil.scala
```diff
@@ -83,7 +83,9 @@ object JDBCUtil {
               }
               if (c == ';' && !stringLiteral) {
                 val sql = new String(out.toByteArray, "UTF-8")
-                conn.update(sql.trim)
+                if(sql != null && !sql.isEmpty()) {
```
takezoe, line 86:
```suggestion
                if (sql != null && !sql.isEmpty()) {
```

takezoe: Maybe this will solve the format error.

src/main/scala/gitbucket/core/util/JDBCUtil.scala
```diff
@@ -83,7 +83,9 @@ object JDBCUtil {
               }
               if (c == ';' && !stringLiteral) {
                 val sql = new String(out.toByteArray, "UTF-8")
-                conn.update(sql.trim)
+                if (sql != null && !sql.isEmpty()) {
+                  conn.update(sql.trim)
```
takezoe, line 87:
```suggestion
                  conn.update(sql.trim)
```

Issue #2628, opened 2021-01-25
Internal Server Error when delete this repository


s-takatsu:
## Issue
**Impacted version**: 4.35.3
**os version**: windows 10 pro, windows server 2016 both

**Deployment mode**: use GitBucket : standalone app,

**Problem description**:
when delete this repository , occured

*****
Internal Server Error
org.apache.commons.io.IOExceptionList: 1 exceptions: [java.io.IOException: Unable to delete file: c:\opt\gitbucket\.gitbucket\repositories\root\hoge.wiki.git\objects]
*****

shreejanshrestha:
Same issue is appearing probably from 4.34.0.
Error log below:
-
Internal Server Error

org.apache.commons.io.IOExceptionList: 1 exceptions: [java.io.IOException: Unable to delete file: f:\gitbucket\tmp\20210131\.\home\repositories\root\test0.wiki.git\objects]

org.apache.commons.io.FileUtils.cleanDirectory(FileUtils.java:345)
org.apache.commons.io.FileUtils.deleteDirectory(FileUtils.java:1206)
gitbucket.core.service.RepositoryService.$anonfun$deleteRepository$1(RepositoryService.scala:241)
scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
gitbucket.core.util.LockUtil$.$anonfun$lock$1(LockUtil.scala:30)
gitbucket.core.util.SyntaxSugars$.defining(SyntaxSugars.scala:12)
gitbucket.core.util.LockUtil$.lock(LockUtil.scala:27)
gitbucket.core.service.RepositoryService.deleteRepository(RepositoryService.scala:237)
gitbucket.core.service.RepositoryService.deleteRepository$(RepositoryService.scala:236)
gitbucket.core.controller.RepositorySettingsController.deleteRepository(RepositorySettingsController.scala:26)
gitbucket.core.controller.RepositorySettingsControllerBase.$anonfun$$init$$51(RepositorySettingsController.scala:433)
gitbucket.core.util.OwnerAuthenticator.$anonfun$authenticate$1(Authenticator.scala:38)
scala.Option.map(Option.scala:242)
gitbucket.core.util.OwnerAuthenticator.authenticate(Authenticator.scala:36)
gitbucket.core.util.OwnerAuthenticator.ownerOnly(Authenticator.scala:30)
[...]

PR #2711, opened 2021-03-20
Fix issue 2628

shreejanshrestha:
●Added function to delete given folder recursively.
●Used the function for deleting the folder, while deleting repository.

fixes gitbucket/gitbucket#2628


takezoe:
Thanks for creating this pull request. Let's see how this works!

PR #2713, opened 2021-03-22
Support HTTP basic authentication

c0llab0rat0r:
Support HTTP basic authentication, to allow api access on non-public instances of GitBucket.

This change is required to support accessing an H2 plugin generated backup with `curl` when an installation of GitBucket is set to disallow anonymous access to the entire site.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/ControllerBase.scala
+++ b/src/main/scala/gitbucket/core/controller/ControllerBase.scala
@@ -7,6 +7,7 @@ import gitbucket.core.model.Account
 import gitbucket.core.service.{AccountService, RepositoryService, SystemSettingsService}
 import gitbucket.core.util.SyntaxSugars._
 import gitbucket.core.util.Directory._
+import gitbucket.core.util.HttpCredentialParser.parseBasicAuth
 import gitbucket.core.util.Implicits._
 import gitbucket.core.util._
 import org.json4s._
@@ -35,6 +36,7 @@ import org.json4s.Formats
  */
 abstract class ControllerBase
     extends ScalatraFilter
+    with AccountService
     with ValidationSupport
     with JacksonJsonSupport
     with I18nSupport
@@ -93,8 +95,16 @@ abstract class ControllerBase
     }
   }
 
+  private def RequestAccount: Option[Account] =
+    Option
+      .apply(request.getHeader("Authorization"))
+      .flatMap(h => parseBasicAuth(h))
+      .flatMap(ba => authenticate(loadSystemSettings(), ba.userName, ba.password))
+
   private def LoginAccount: Option[Account] =
-    request.getAs[Account](Keys.Session.LoginAccount).orElse(session.getAs[Account](Keys.Session.LoginAccount))
+    RequestAccount
+      .orElse(request.getAs[Account](Keys.Session.LoginAccount))
+      .orElse(session.getAs[Account](Keys.Session.LoginAccount))
 
   def ajaxGet(path: String)(action: => Any): Route =
     super.get(path) {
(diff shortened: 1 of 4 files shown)


takezoe:
Thank you for sending this pull request!

We need to be careful whenever we introduce this kind of system-wide feature that is difficult to change once implemented.

In my personal opinion, applying [ApiAuthenticationFilter](https://github.com/gitbucket/gitbucket/blob/master/src/main/scala/gitbucket/core/servlet/ApiAuthenticationFilter.scala) to all requests might be easier and more consistent because it also supports API token. Using API token for automation is generally safer than using a user's password.

Also, `ApiAuthenticationFilter` implementation can be improved. For example, parsing `Authorization` header can be separated from the filter so that we can write tests as you did in this pull request.

Anyway, I'm still considering what is the best way. Does anyone have options or feedbacks about this?

kounoike:
My opinion:
That's not a good enough reason to accept that change. Because h2-backup-plugin has API endpoint. You can trigger backup by API token based authentication.
https://github.com/gitbucket-plugins/gitbucket-h2-backup-plugin/blob/ce5f55649b24a646f38acdd3dd06168c7873d0f9/src/main/scala/fr/brouillard/gitbucket/h2/controller/H2BackupController.scala#L44
(But, it's not documented in README.md)

c0llab0rat0r:
Once I've generated an api token, how do I use it to trigger the backup?

takezoe:
You can use API token like this: https://github.com/gitbucket/gitbucket/wiki/API-WebHook#example

c0llab0rat0r:
When I try to use an api token for the root account, I get `401 Unauthorized`:

    curl -i -H "Authorization: token somelonghash" https://gitbucket.somedomain/database/backup

...responds with:

    HTTP/1.1 401 Unauthorized
    Server: nginx/1.17.10 (Ubuntu)
    Date: Sun, 04 Apr 2021 02:46:05 GMT
    Content-Type: text/html;charset=utf-8
    Content-Length: 0
    Connection: keep-alive
    Set-Cookie: JSESSIONID=thesessionid; Path=/; secure; Secure; HttpOnly
    Expires: Thu, 01 Jan 1970 00:00:00 GMT

c0llab0rat0r:
This doesn't include `/database/backup` since it's not underneath `/api`:

    context.addFilter("apiAuthenticationFilter", new ApiAuthenticationFilter)
    context
      .getFilterRegistration("apiAuthenticationFilter")
      .addMappingForUrlPatterns(EnumSet.allOf(classOf[DispatcherType]), true, "/api/*")

Is that why my `curl` request with `Authorization: token` fails against `/database/backup` ?

Does that mean that the url mapping for `apiAuthenticationFilter` needs to include `/database` ?

takezoe:
Why don’t you use /api/v3/plugins/database/backup?

c0llab0rat0r:
> Why don’t you use `/api/v3/plugins/database/backup`?

It looks like I can. I've submitted a documentation PR just now (https://github.com/gitbucket-plugins/gitbucket-h2-backup-plugin/pull/32) to indicate using the `/api/v3/plugins/database/backup` endpoint instead of the `/database/backup` endpoint.

takezoe:
Thank you!

c0llab0rat0r:
I'm amicable towards my pr getting canned since my use case has been solved via the api tokens on the api url.

I wouldn't mind changing it to a refactor either, or leaving it open for further discussion and repurposing.

takezoe:
Thank you! I tried to reuse codes and test cases in this pull request to refactor `ApiAuthenticationFilter` and other codes that refer `Authorization` header. However, it was harder than expected because each one has different requirements.

takezoe:
For now, I would close this as won't fix. Thank you for your contributions to gitbucket-h2-backup-plugin!

PR #2732, opened 2021-04-07
(bugfix) password msg when editing an LDAP profile

### LINKED ISSUE #2669: LDAP profile edit has misleading confirmation prompt
## Issue
**Impacted version**: 4.35.3

**Deployment mode** : reproduced with a standalone app and deployed with Apache Tomcat

**Problem description**:
To reproduce, have LDAP authentication enabled, as documented at https://github.com/gitbucket/gitbucket/wiki/LDAP-Authentication-Settings
Log in with a user that is authenticated through that LDAP, and edit your profile (or, as in my experience, be forced to update email upon first sign-on). 

When editing the profile (`/profile/_edit`) for an account created via LDAP, we always get falsely prompted, "Are you sure you want to change password?" when clicking "Save". 
This is of course misleading and confusing for users since they don't have a password field and don't wish to change their password on the underlying LDAP system. 

After some inspection of the web page, it appears the problem is stemming from here:
https://github.com/gitbucket/gitbucket/blob/cbddc34bfa3a994912d1db232d70a90ae6fbfc9e/src/main/twirl/gitbucket/core/account/edit.scala.html#L75
and the fact that the password field is conditionally excluded from this HTML panel for LDAP accounts. The check needs to account for the possibly-not-there password field. Perhaps something like this (untested) JavaScript:
```javascript
    let pwval = $('#password').val();
    if(typeof(pwval) != 'undefined' && pwval != ''){
      return confirm('Are you sure you want to change password?');
    } else {
      return true;
    }
```

ThePrez:

Closes #2669.
Opening as draft until tests and compilation are confirmed

### DIFF
--- a/src/main/twirl/gitbucket/core/account/edit.scala.html
+++ b/src/main/twirl/gitbucket/core/account/edit.scala.html
@@ -72,7 +72,8 @@ $(function(){
   addExtraMailAddress();
   $('#extraMailAddresses').on('change', '.extraMailAddress', checkExtraMailAddress);
   $('#save').click(function(){
-    if($('#password').val() != ''){
+    let pwval = $('#password').val();
+    if(typeof(pwval) != 'undefined' && pwval != ''){
       return confirm('Are you sure you want to change password?');
     } else {
       return true;


ThePrez:
Ready for review. Thanks!

takezoe (approved):
Looks good. Thank you!

Issue #2506, opened 2020-08-07
The GitHub plugin in Jenkins does not detect tags


ziggystar:
## Issue
**Impacted version**: 4.34.0

**Deployment mode**: war in tomcat

**Problem description**:
I have setup our GitBucket as a GitHubEnterprise server in Jenkins. I configure a GitHub-Organization Job that scans all repos within an org. This works. But I cannot get Jenkins to detect tags. When detection of tags is enabled, the output of the scan log is something like this:

```
  1 branches were processed
  Checking pull-requests...
  0 pull requests were processed
  Checking tags...
  Getting remote tags...
  0 tags were processed
```
I have filed previous issues, which targeted the problematic part of the API implementation. I see that this appears to be a moving target, as also the library that is used within the Jenkins GitHub plugin is currently working on this part of their code base.

Original Issue with more details, but might be outdated meanwhile:  #2437
Also related:  #2461

I suggest to leave this issue open until detection of tags in Jenkins works. No matter what the actual problem is.

PR #2807, opened 2021-06-28
Fix refs API as far as Jenkins github-branch-source plugin can detect tags

ziggystar:

This PR adresses #2506, though I'm not sure whether it is now completely fixed. So I suggest leaving the issue open. 

What I did and achieved:

I added (and renamed) fields for the GHRef objects that are returned on `https://api.github.com/repos/<user>/<repo>git/refs/tags`. I also had to fix the routing to accept the `.../refs/tags` URL.

Here is an example what GH returns under [https://api.github.com/repos/gitbucket/gitbucket/git/refs/tags/](https://api.github.com/repos/gitbucket/gitbucket/git/refs/tags/):

The `node_id` is still missing. Also I'm not so sure about what I put under the URLs. And further I did not remove any of the existing fields (though I renamed some to match the GH API). Please have a look here, particularly about the URLs.

```
[
  {
    "ref": "refs/tags/1.0",
    "node_id": "MDM6UmVmOTM1MDc0NjpyZWZzL3RhZ3MvMS4w",
    "url": "https://api.github.com/repos/gitbucket/gitbucket/git/refs/tags/1.0",
    "object": {
      "sha": "1f164ecf2f59190afc8d7204a221c739e707df4c",
      "type": "tag",
      "url": "https://api.github.com/repos/gitbucket/gitbucket/git/tags/1f164ecf2f59190afc8d7204a221c739e707df4c"
    }
  },
  {
    "ref": "refs/tags/1.1",
    "node_id": "MDM6UmVmOTM1MDc0NjpyZWZzL3RhZ3MvMS4x",
    "url": "https://api.github.com/repos/gitbucket/gitbucket/git/refs/tags/1.1",
    "object": {
      "sha": "8aec16bf7f164e99605a26b337a79246ca374303",
      "type": "tag",
      "url": "https://api.github.com/repos/gitbucket/gitbucket/git/tags/8aec16bf7f164e99605a26b337a79246ca374303"
    }
  },
  {
    "ref": "refs/tags/1.2",
    "node_id": "MDM6UmVmOTM1MDc0NjpyZWZzL3RhZ3MvMS4y",
    "url": "https://api.github.com/repos/gitbucket/gitbucket/git/refs/tags/1.2",
    "object": {
      "sha": "f9cf451ceabdc9bd422ce6ac24b1c2aa397a00c2",
      "type": "tag",
      "url": "https://api.github.com/repos/gitbucket/gitbucket/git/tags/f9cf451ceabdc9bd422ce6ac24b1c2aa397a00c2"
    }
  },
  …
]
```

Using a fresh install of Jenkins, the gh-branch-plugin can now successfully detect tags.


takezoe (approved):
@ziggystar Modifying the existing test cases entirely based on the response (adding missing fields and removing unnecessary fields) would be fine. Could you update the pull request?

ziggystar:
@takezoe I just had a look at the tests and at what the API returns. My change was just about the `ref/tags` call, but it appears that all `ref/*` calls should be fixed in the same way.

I'll try to do this in the following day.

Do you know anything about this `node_id`? This is something about GraphQL. I suppose GB does not have this feature, so should we add the field as an empty string?

takezoe:
At GitHub, `node_id` can be used for identifying objects of REST API in GraphQL API. It's described in the following document in detail. I don't know how is `node_id` generated, though.
https://docs.github.com/en/graphql/guides/using-global-node-ids

Anyway, GitBucket doesn't have GraphQL API for now, so a dummy value should be fine. Rather, we might not need to generate `node_id` field.

ziggystar:
@takezoe Concerning the URLs: There appear to be two different types of refs: "commit" and "tag".
Within the response to the `ref` API-calls, they have URLs pointing to different parts of the API.

For type "commit":

https://api.github.com/repos/gitbucket/gitbucket/git/commits/7deea43c75e976e64df4d0a4490aeefedd9b3207

For type "tag":

https://api.github.com/repos/gitbucket/gitbucket/git/tags/1f164ecf2f59190afc8d7204a221c739e707df4c

Is it correct that GB does currently not implement both APIs? What do we want to put there instead?

ziggystar:
@takezoe I refactored the RefAPI file, and I want to add some integration tests. Is it correct that the integration tests currently do not integrate the http server (URL param extraction)? Or are there such tests in different files, and `ApiIntegrationTests.scala` could be extended to cover also the URL parsing part?

ziggystar:
I'm having trouble with the integration tests picking up my source changes. Even if I do a clean recompile, the tests run against the old code. 

Sometimes they start hanging and do not run at all.

This is happening when running tests in Idea or directly from SBT. Sometimes I kill processes, delete `target` directory, and then it works again once (maybe twice). 

Any ideas?

takezoe:
@ziggystar Integration tests require `sbt package` because it works against a created war file.
https://github.com/gitbucket/gitbucket/blob/e74db57b750e3591afecfcd77e0a08972664191c/build.sbt#L138-L142

takezoe:
@ziggystar Hi! Would you add integration tests for this? Or shall I merge this without tests for now?

ziggystar:
I've pushed my current state. I didn't get back to this, and I'm now on vacation for two weeks.

Problem is that I started refactoring the code on a larger scale. I think the code is good. I had unit tests passing with two different samples from GitHub.

 I started working on the integration tests. I'm not sure those are done or meaningful. I think the refactoring should have at least one/two integration tests.

takezoe:
Oh, I see. We are not in a hurry for this. Please let me know if you can make any progress or if you have any problems. Once merging this and refactoring in another PR could be an option :-)

takezoe:
I would take over this pull request to include this in the next release.

PR #2896, opened 2021-10-13
Add HTTPS support for JettyLauncher

imeszaros:

### DIFF
--- a/src/main/java/JettyLauncher.java
+++ b/src/main/java/JettyLauncher.java
@@ -6,6 +6,8 @@ import org.eclipse.jetty.server.SecureRequestCustomizer;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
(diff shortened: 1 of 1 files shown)


takezoe:
Thank you! This looks good to me.

By the way, I think it would be great if we can additionally configure the embedded jetty to redirect HTTP to HTTPS when we enable HTTPS (ref: https://stackoverflow.com/questions/26123604/redirect-from-http-to-https-automatically-in-embedded-jetty). In that case, additional configuration would be necessary like `--connectors=http,https --redirect_https=true`. What do you think about this?

takezoe:
Note: we also have to update Wiki for CLI options: https://github.com/gitbucket/gitbucket/wiki/Basic-configurations

takezoe:
@imeszaros  Directly updating Wiki pages is okay. We allow editing Wiki pages for everyone who has a GitHub account.

imeszaros:
@takezoe Oh i see! That was exactly my thouht that the wiki should not be updated before the next version gets released. But with the version annotations it's ok!

Issue #2983, opened 2022-01-30
In issue comments, Markdown tables are displayed overflowing on the right side if the width of the cell is too long


gemmaro:
## Issue
**Impacted version**: 4.37.2 (Earlier versions may also apply, but I haven't checked them all)

**Deployment mode**: Activated as standalone. I also use Nginx as a reverse proxy.

**Problem description**: In issue comments, Markdown tables are displayed overflowing on the right side if the width of the cell is too long.
Look at the following example images:

![Screenshot 2022-01-30 at 10-29-13 test - Issue #1 - root test-ci](https://user-images.githubusercontent.com/29575029/151683583-e1905333-ae30-4aaf-93cd-1326afefb969.png)

**How to reproduce**: For example, write the following Markdown in the comments of an issue, save it, and adjust the width of the browser.

```markdown
| a |
| - |
| bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb |
```

**Current workaround**: If GitBucket version is 4.34.0 or above, by adding the following line to [user-defined CSS](https://gitbucket.github.io/gitbucket-news/gitbucket/2020/07/26/gitbucket-4.34.0.html#user-defined-css), the display will look like the image below, avoiding overlapping due to overflow.

```css
.markdown-body { overflow: auto }
```

![Screenshot 2022-01-30 at 10-48-23 test - Issue #1 - root test-ci](https://user-images.githubusercontent.com/29575029/151683637-6ed9fc42-e3bb-409c-b3f3-17c77dd47b78.png)

zainaftab44:
Fixed this issue in PR #3082. Needed only a little CSS change

PR #3082, opened 2022-07-03
#2983 fixed css for table

zainaftab44:

### DIFF
--- a/src/main/webapp/assets/common/css/gitbucket.css
+++ b/src/main/webapp/assets/common/css/gitbucket.css
@@ -1526,7 +1526,6 @@ div.markdown-body code {
 }
 
 div.markdown-body table {
-  /*width: 100%;*/
   table-layout:fixed;
   width:100%;
   margin-bottom: 20px;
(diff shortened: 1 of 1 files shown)


takezoe:
This will change the presentation of existing tables, but I don't come up with a better idea since I'm not very good at CSS.

src/main/webapp/assets/common/css/gitbucket.css
```diff
@@ -1527,6 +1527,8 @@ div.markdown-body code {
 
 div.markdown-body table {
   /*width: 100%;*/
```
takezoe, line 1529: This comment should be removed if we would add `width:100%;`.

takezoe:
How about using the following styles?
```css
max-width: 100%;
overflow: auto;
display: block;
```
This will make horizontal scrolling only when a table exceeds the parent container size while keeping the existing table presentation.

takezoe:
I would once merge this and maybe improve later as I comment above.

Issue #3434, opened 2023-11-27
Performance problems with requests to `/api/v3/repos/:owner/:repository/branches`


ziggystar:
## Issue
**Impacted version**: 4.39.0

**Deployment mode**: *tomcat*
Requests to  `/api/v3/repos/:owner/:repository/branches` are very slow for repositories that have stale branches.
Those are branches with many commits behind master (thousands). 

The problem is also visible in the loading of the branch view of a repo.

I think I tracked it down to the problem that loading the branches also walks through commits to find the number if diverging commits.

This takes 10s of seconds for our repo. This also results in Jenkins repo scan for new branches and tags failing with a timeout.

A solution probably needs to either make the rev walking faster or at least don't do it for the API request.

PR #3436, opened 2023-11-27
fix: add GitUtils.findBranchesNoMergeInfo and use it when merge info …

ziggystar:
Creates a new method `JGitUtil.getBranchesNoMergeInfo` that does not perform the expensive repo walks. Then use this method in API requests to list branches.

This is a partial fix for #3434 which improves performance for the API requests. Rendering of the branches page of the repo is still slow.

Maybe renaming the existing method `getBranches` to `getBranchesWithMergeInfo` instead would be a better approach, since the new method should be used whenever possible.

### DIFF
--- a/src/main/scala/gitbucket/core/util/JGitUtil.scala
+++ b/src/main/scala/gitbucket/core/util/JGitUtil.scala
@@ -248,6 +248,14 @@ object JGitUtil {
 
   case class BranchMergeInfo(ahead: Int, behind: Int, isMerged: Boolean)
 
+  case class BranchInfoSimple(
+    name: String,
+    committerName: String,
+    commitTime: Date,
+    committerEmailAddress: String,
+    commitId: String
+  )
+
   case class BranchInfo(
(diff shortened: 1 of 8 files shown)


src/main/scala/gitbucket/core/util/JGitUtil.scala
```diff
@@ -1244,6 +1252,25 @@ object JGitUtil {
     git.log.add(startCommit).addPath(path).setMaxCount(1).call.iterator.next
   }
 
+  def getBranchesNoMergeInfo(git: Git, defaultBranch: String): Seq[BranchInfoSimple] = {
```
takezoe, line 1255: Can we add a test case for this method? The same test case as `getBranches` would be enough.
https://github.com/gitbucket/gitbucket/blob/9caa73a5854ca4c5c32a27a1677cc39859428425/src/test/scala/gitbucket/core/util/JGitUtilSpec.scala#L147-L174

PR #3647, opened 2024-11-29
Calculate diffs between commits based on common ancestor; works …

ziggystar:
…for force push

I've rewritten `JGitUtil.getCommitLog` to properly handle merge bases. It probably also is faster in general, since it uses the proper JGit tools. 
The previous code returned all previous commits in cases of force push (since the old id is gone). The new code now uses JGit to find the merge base with the old id.

**This changes behavior**
I am totally not sure how this method should behave at all. We've been running this code for some months on our instance and it worked.

I've added a test so the new code gets executed. But it's just a comparison with old method for a simple case. Since the behavior differs I cannot compare for a more complex case. Also I don't know the intended contract of `getCommitLog`, so I can't write a proper test.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/ReleasesController.scala
+++ b/src/main/scala/gitbucket/core/controller/ReleasesController.scala
@@ -130,11 +130,10 @@ trait ReleaseControllerBase extends ControllerBase {
   })
 
   get("/:owner/:repository/changelog/*...*")(writableUsersOnly { repository =>
-    val Seq(previousTag, currentTag) = multiParams("splat")
-    val previousTagId = repository.tags.collectFirst { case x if x.name == previousTag => x.commitId }.getOrElse("")
-
     val commitLog = Using.resource(Git.open(getRepositoryDir(repository.owner, repository.name))) { git =>
-      val commits = JGitUtil.getCommitLog(git, previousTagId, currentTag).reverse
+      val Seq(previousTag, currentTag) = multiParams("splat")
+
+      val commits = JGitUtil.getCommitLog(git, previousTag, currentTag).reverse
       commits
         .map { commit =>
           s"- ${commit.shortMessage} ${commit.id}"
--- a/src/main/scala/gitbucket/core/util/JGitUtil.scala
+++ b/src/main/scala/gitbucket/core/util/JGitUtil.scala
@@ -28,6 +28,7 @@ import org.eclipse.jgit.util.io.DisabledOutputStream
 import org.slf4j.LoggerFactory
 
 import scala.util.Using.Releasable
+import scala.util.{Try, Using}
 
 /**
  * Provides complex JGit operations.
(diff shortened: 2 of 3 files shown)


takezoe:
Thank you! This looks pretty good.

takezoe:
This doesn't work for tags.
```
org.eclipse.jgit.errors.IncorrectObjectTypeException: Object c909bc4b19f02688adc479c6e42000539476112e is not a commit.
        at org.eclipse.jgit.internal.storage.file.WindowCursor.open(WindowCursor.java:141)
        at org.eclipse.jgit.revwalk.RevWalk.getCachedBytes(RevWalk.java:1188)
        at org.eclipse.jgit.revwalk.RevCommit.parseHeaders(RevCommit.java:149)
        at org.eclipse.jgit.revwalk.RevWalk.markStart(RevWalk.java:349)
        at org.eclipse.jgit.api.LogCommand.add(LogCommand.java:341)
        at org.eclipse.jgit.api.LogCommand.add(LogCommand.java:178)
        at gitbucket.core.util.JGitUtil$.$anonfun$getCommitLog$7(JGitUtil.scala:682)
        at scala.Option.flatMap(Option.scala:283)
        at gitbucket.core.util.JGitUtil$.getCommitLog(JGitUtil.scala:679)
        at gitbucket.core.controller.ReleaseControllerBase.$anonfun$$init$$26(ReleasesController.scala:136)
        at scala.util.Using$.resource(Using.scala:296)
        at gitbucket.core.controller.ReleaseControllerBase.$anonfun$$init$$25(ReleasesController.scala:133)
        ...
```

src/main/scala/gitbucket/core/util/JGitUtil.scala
```diff
@@ -659,17 +660,44 @@ object JGitUtil {
...
    *
    * @param git the Git object
    * @param from the from revision
    * @param to the to revision
    * @return the commit list
    */
   // TODO swap parameters 'from' and 'to'!?
-  def getCommitLog(git: Git, from: String, to: String): List[CommitInfo] =
-    getCommitLogs(git, to)(_.getName == from)
+  def getCommitLog(git: Git, from: ObjectId, to: ObjectId): List[CommitInfo] =
```
takezoe, line 678: It seems that `git.getRepository.parseCommit()` needs to be called with `from` and `to` to support tags.

ziggystar:
I've updated `getCommitLog(Git, String, String)` to resolve tag ObjectIds using `RefDatabase.peel`.

I'm really not sure about what `getCommitLog(Git, String, String)` must accept. I feel that this logic belongs to the calling site. It's only used at GitRepositoryServlet.scala:501 . The other usage is at line 299, where we should already have the ObjectIds (hopefully no tag objects) and can call `getCommitLog(Git,ObjectId,ObjectId)` directly.

src/main/scala/gitbucket/core/controller/ReleasesController.scala
```diff
@@ -130,11 +130,10 @@ trait ReleaseControllerBase extends ControllerBase {
...
 
   get("/:owner/:repository/changelog/*...*")(writableUsersOnly { repository =>
-    val Seq(previousTag, currentTag) = multiParams("splat")
-    val previousTagId = repository.tags.collectFirst { case x if x.name == previousTag => x.commitId }.getOrElse("")
-
     val commitLog = Using.resource(Git.open(getRepositoryDir(repository.owner, repository.name))) { git =>
-      val commits = JGitUtil.getCommitLog(git, previousTagId, currentTag).reverse
+      val Seq(previousTag, currentTag) = multiParams("splat").map(git.getRepository.resolve)
+
+      val commits = JGitUtil.getCommitLog(git, previousTag, currentTag).reverse
```
takezoe, line 136: This should call `getCommitLog(git: Git, from: String, to: String)`

src/main/scala/gitbucket/core/util/JGitUtil.scala
```diff
@@ -659,17 +660,55 @@ object JGitUtil {
...
+      }
+    }
+
+    getCommitLog(git, resolveString(from), resolveString(to))
+  }
+
   /**
    * Returns the commit list between two revisions.
+   * This method takes strings as parameters, but does not define what happens when the strings are no valid
+   * commit ids. They can be either invalid syntactically, or they can be not within the repo.
```
takezoe, line 681: Just saying from and to must be valid commit ids would be enough to clarify expectation.

takezoe:
Sorry for the current messy implementation 🙇‍♂️

With your latest fix, `JGitUtil` now has new `getCommitLog(git: Git, from: String, to: String)` with compatibility with the current one at least for known usages which is great.

One last thing is that `ReleasesController` should call `getCommitLog(git: Git, from: String, to: String)` because tags are expected to be given there (Of course, it would work even for commits and branches though).

Screenshot 2024-12-03 at 21 42 08

ziggystar:
I've pointed `ReleasesController.getCommitLog` to the String-based method and improved some of the javadoc. Sorry, I think I did not understand your previous comment.

Side-note: I tried to write a test for the changelog endpoint. But I did not find any such tests that make or simulate the final http requests. Is this correct?

takezoe:
> Side-note: I tried to write a test for the changelog endpoint. But I did not find any such tests that make or simulate the final http requests. Is this correct?

Yes. We only have such test cases for Web APIs.

takezoe (approved):
LGTM

takezoe:
Thank you for great work! I wonder if we can replace other `getCommitLog()` methods too.

PR #3653, opened 2024-12-06
fix: make some methods call JGitUtil.getCommitLog with ObjectId

ziggystar:

This is a fix for the PR #3647 .

It seems like the change with the tag resolution broke something. There are occasions where `JGit.getCommitLog(Git,String,String)` gets called with invalid names. This happens when during a push there is no old id. Then the names will be just zeros `000000...` (probably from the zero `ObjectId`). This currently cannot be resolved. 

I fixed it by using the `getCommitLog(Git,ObjectId,ObjectId)` version directly, which also makes more sense here.

I think the behavior in this case (maybe like it was before), is that the full log will be returned, since we do not have a `from` (or the from will be filtered out). I'm not so sure that this is what is intended in `GitRepositoryServlet`. I think the intention is to find only the new commits and do some processing with them (Wiki, Issues). But shouldn't this only happen for the main branch?

### DIFF
--- a/src/main/scala/gitbucket/core/servlet/GitRepositoryServlet.scala
+++ b/src/main/scala/gitbucket/core/servlet/GitRepositoryServlet.scala
@@ -296,7 +296,7 @@ class CommitLogHook(owner: String, repository: String, pusher: String, baseUrl: 
     } else {
       command.getType match {
         case ReceiveCommand.Type.DELETE => Nil
-        case _                          => JGitUtil.getCommitLog(git, command.getOldId.name, command.getNewId.name)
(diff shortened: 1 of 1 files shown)


takezoe:
> shouldn't this only happen for the main branch?

There are various things handled in commit hook.

- Record activity logs
- Create links to other issues mentioned in commit message
- Call web hooks
- Call plugin's hooks

Those things should happen even on branches. Otherwise, even CI will not run on branches.

takezoe:
Using `ObjectId` version makes sense. Or we may add zeroId handling in `String` version.

takezoe (approved):
LGTM

PR #3899, opened 2025-12-23
Remove text-align: left from the cells of a table created with markdown

yasumichi:
I think it is unnecessary to set text-align in the cells of a table created with markdown.


takezoe (approved):
`text-align: left` is default for `td` and Bootstrap has `text-align: left` for `th`. So, this style definition is redundant. It's fine to remove.

On the other hand, there are many redundant styles in `gitbucket.css`. For example, not only `text-align: left` but also
```css
  vertical-align: top;
  border-top: 1px solid #dddddd;
```
seem to be unnecessary here. Maybe it's better to fully refresh `gitbucket.css`.

PR #3996, opened 2026-03-28
Improve pull request compare performance by making mergeability and diff loading manually triggerable, with configurable default mode

shreejanshrestha:

* although no issuses are going to be solved by this PR, have marked 【 [marked as closed using commit message](https://help.github.com/articles/closing-issues-via-commit-messages/) all issue ID that this PR should correct】 as checked. sorry if I misunderstood.

Impacted version:
master

Deployment mode:
Development build from source (standalone app)

Problem description:
When opening the pull request compare screen for branches with many commits/files, the page can feel slow because expensive compare work and mergeability checks run automatically.
This creates a poor experience in large repositories and makes users wait before they can interact with the page.

Steps to reproduce:

- Open a repository with many commits/files between base and compare branches.
- Click New pull request.
- Observe initial compare page load time.
- Change base/compare dropdowns and observe repeated loading.

Current pain points:

- Full compare computation can run before the user explicitly asks for it.
- Mergeability checks can run automatically in cases where users prefer manual control.
- Performance behavior is not always predictable from the UI.

Implemented behavior:

- Added Check mergeability button to explicitly trigger mergeability checking and full compare flow.
- Added quick/no-check compare path to avoid expensive work on initial load.
- Added global admin setting to choose default mode:
  - Check by default (※means when compare is display it will trigger load diff/ check mergable)
  - No check by default (fast/manual)

Related commits:

- 53730c6f9 disable auto-checking of conflict and diff loading, when creating new pull request.
- 5ade95503 make auto check(ON/OFF) configurable from setting.

Expected result:

- Admin can choose default behavior globally based on instance size and user expectations.
- Users can still explicitly trigger mergeability check from the compare screen.
- Large repositories get faster, more predictable PR compare interactions.

Notes:

Global default is "Check by default" for upgrade compatibility.

### DIFF
--- a/src/main/scala/gitbucket/core/controller/PullRequestsController.scala
+++ b/src/main/scala/gitbucket/core/controller/PullRequestsController.scala
@@ -375,6 +375,8 @@ trait PullRequestsControllerBase extends ControllerBase {
 
   get("/:owner/:repository/compare")(referrersOnly { forkedRepository =>
     val headBranch = params.get("head")
+    val quickLoad = params.get("quick").contains("true")
+    val quickQuery = if (quickLoad) "?quick=true" else ""
     (forkedRepository.repository.originUserName, forkedRepository.repository.originRepositoryName) match {
       case (Some(originUserName), Some(originRepositoryName)) =>
         getRepository(originUserName, originRepositoryName).map { originRepository =>
@@ -388,7 +390,7 @@ trait PullRequestsControllerBase extends ControllerBase {
               .getOrElse(JGitUtil.getDefaultBranch(oldGit, originRepository).get._2)
 
             redirect(
-              s"/${forkedRepository.owner}/${forkedRepository.name}/compare/$originUserName:$oldBranch...$newBranch"
+              s"/${forkedRepository.owner}/${forkedRepository.name}/compare/${originUserName}:${oldBranch}...${newBranch}${quickQuery}"
             )
           }
         } getOrElse NotFound()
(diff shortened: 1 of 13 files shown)


src/main/scala/gitbucket/core/service/SystemSettingsService.scala
```diff
@@ -281,6 +283,7 @@ object SystemSettingsService {
     gravatar: Boolean,
     notification: Boolean,
     limitVisibleRepositories: Boolean,
+    compareNoCheckByDefault: Boolean = false,
```
takezoe: To align with all other properties, no default value is preferrable. In this case, we need to update test cases though.

takezoe (changes requested):

src/main/scala/gitbucket/core/service/SystemSettingsService.scala
shreejanshrestha: thank you for the comment.
implemented by giving a default value of false, so as to maintain compatibility with plugins, built against "gitbucket.core.service". i cannot verify all the plugins, so i think its little bit difficult to align with other properites. what do you say? i may be wrong as i don't know how plugins are begin developed and released. could you suggest?

takezoe: `SystemSettings` is expected to load using `loadSystemSettings()` and accessed via `Context`. Simply removing the default value from the case class and providing it in `loadSystemSettings()` (as you've done) is fine.

shreejanshrestha: Thank you for the explanation.🙇‍♂️
Removed the default.

src/main/scala/gitbucket/core/controller/PullRequestsController.scala
```diff
@@ -436,76 +441,113 @@ trait PullRequestsControllerBase extends ControllerBase {
     val Seq(origin, forked) = multiParams("splat")
     val (originOwner, originId) = parseCompareIdentifier(origin, forkedRepository.owner)
     val (forkedOwner, forkedId) = parseCompareIdentifier(forked, forkedRepository.owner)
+    val requestedCheck = params.get("check").contains("true")
+    val quickLoad = params
+      .get("quick")
+      .map(_.equalsIgnoreCase("true"))
+      .getOrElse(!requestedCheck && context.settings.basicBehavior.compareNoCheckByDefault)
+    val autoMergecheck = requestedCheck && !quickLoad
```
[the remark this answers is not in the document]
shreejanshrestha: Thank you.
Removed the dead code.👍

takezoe (approved):
LGTM. As a futher improvement, we could provide an option to enable this feature by the number of commits in the diff.

shreejanshrestha:
takezoe san,
thanks for merging and new improvement idea. was thinking of providing checkByDefault setting per repository, but your suggestion also seems good. will try it after somedays.👍

PR #4072, opened 2026-07-12
Add GitHub API Compatibility for Repository ID Numbering And Managing Forks

iamjpotts:

# Overview

## Adds Additional GitHub API Compatibility

* Add id number to repositories, and support it on the GitHub compatibility API
* Add fork flag field to GitHub compatibility API
* Add additional routes to the GitHub compatibility API:
  * `GET /api/v3/repositories/:id`
  * `POST /api/v3/repos/:owner/:repository/forks`
  * `GET /api/v3/repos/:owner/:repository/forks`

## Fixes Two Existing Bugs 

* Fix: for invalid parent and origin repository references, add the missing user(s) and repositories(s); then add a constraint to prevent this
* Fix: preserve issue assignees when renaming a repository

## Maintenance

* Refactor: Simplify repository rename logic via cascading constraints (the need to recreate the table for id's is an opportunity for this)

## Limitations Of New GitHub Compatibility

* Creating a fork could be asynchronous but is synchronous for now; it returns 202 Accepted for api compatibility
* New compatibility API for listing forks supports sorting but not by stargazers or watchers (returns 501 Not Implemented)

# Testing

* New automated tests
* Upgrade from backup - I restored a backup of my own instance (around 100 repositories, though not multi user) and spot checked data
* I manually ran `gh` cli commands against the new apis
* A shell script with automated checks using the `gh` cli
* A closed-source Kubernetes operator which creates GitBucket instances and exercises backup/restore and fresh install scenarios

After code review items are worked through (in this pr or a split off pr), I plan to add terminal output showing use of the `gh` cli with the new routes against the final changes net of all feedback.

# Alternatives

The branch grew a bit bigger than I care for, for a single PR. Wanted to get your feedback on the changes overall before potentially breaking it up into multiple prs (unless you'd rather keep it as one).

## Data Preservation Test - Overall Approach
I am not sure I like how the test in the last commit is setup, but it needs to populate and check 19 tables. Open to suggestions if you don't care for its approach. It does seem like an improvement over the 4.35 migration for web_hook, which doesn't appear to have much test coverage. 

## Data Preservation Test - Both Min and Max DB Platform Versions

I am also on the fence as to whether its necessary to test the data preservation on both min and max supported database platform versions, since the migration from start to finish is already doing that. Pro is more coverage, con is longer test runs - the tests using test containers are a bit slow and don't appear to run in parallel.

## Fixing Orphaned Data (Invalid Usernames and Repository Names)

Since the proposed migrations add foreign key constraints to the origin and parent repository references, the migration needs to handle data that doesn't match the account table or repository table. Non-matching data could be added back (which is done here), or it could be erased (ie with a null). The former requires re-creating empty git repositories on disk since a substantial amount of the application logic expects that for a repository row.

## Or Multiple PRs

Here's an idea for how this pr could be broken up:

* 4.46.2 - Fix - preserve issue assignees when renaming a repository
* 4.47 - Additional GitHub API compatibility + resolve invalid parent and origin references

Most of the complexity (and thus pr discussion) will be in that second item.


takezoe:
@iamjpotts Thank you for the great work!

Just wondered if it's possible to replace all foreign keys with `repository id` though it would require massive amount of work. What do you think?

iamjpotts:
> @iamjpotts Thank you for the great work!
> 
> Just wondered if it's possible to replace all foreign keys with `repository id` though it would require massive amount of work. What do you think?

I think that's a natural eventuality, but would be too much additional complexity for this pr.

Something else I noticed while working on this, the default branch column is nullable, but much of the code assumes it is populated. Since this pr is re-creating the repository table, I think the column should be made non-nullable and have a default (probably `main`). Any (seemingly unlikely) rows with a null value would need their value set, which could be the column default. Upside of using the column default is its simple to implement (just a sql insert statement), downside is a repository without a default branch set might have branches but no main or master branch (unusual but possible edge case), which would require iterating over all of the repos on disk at startup when the most recent migration is before 4.47.

What do you want done about the default branch nullability issue - leave the column as-is (nullable), use sql to set it to `main`, or something more sophisticated?

iamjpotts:
I've now reviewed the open issues. There are no open issues related to this pr specifically.

takezoe:
> I think that's a natural eventuality, but would be too much additional complexity for this pr.

OK, it won't happen in this PR. I got it.

> Something else I noticed while working on this, the default branch column is nullable, but much of the code assumes it is populated. Since this pr is re-creating the repository table, I think the column should be made non-nullable and have a default (probably main).

`SystemSettings.defaultBranch` should be used but it can't be defined as a column default value in that case. Also, there are other problems that can't be solved by simply defining a default value. It may be better to consider this more carefully and create a separate PR.

iamjpotts:
> Also, there are other problems that can't be solved by simply defining a default value

Anything come to mind immediately?

takezoe:
As you mentioned, when a default branch doesn't exist. GitBucket allows creating an empty repository which doesn't habe any branch, so it's not a rare case. I think allowing null and handling it properly is a right way to go. Anyway, it should be out of scope for this pull request.

iamjpotts:
Replaced by https://github.com/gitbucket/gitbucket/pull/4083.

* https://github.com/gitbucket/gitbucket/pull/4077 was carved out as a preliminary separate fix
* Substantial simplification to migrations (no more recreating the repository table)

> [the default branch issue] should be out of scope for this pull request.

Agree.

PR #4101, opened 2026-08-09
Improve GitBucketCoreModuleSpec test suite runtime by starting all containers at once

iamjpotts:

Reduces total test suite time (ie sbt test) by about 50 seconds. While the postgres containers start quickly, the MySQL containers take 20-30s to start.

This change starts all of the containers as the tests are registering, and each test awaits its own container when it starts executing. Containers start in parallel; test still execute in serial. The h2 tests execute first in parallel with MySQL containers starting.

A downside of this change is that a targeted/filtered sbt test run (eg for only h2 tests) still starts all of those containers and waits for them all to shut down, which increases the runtime of an h2-only run from 2s to 20s. This could be reduced back to single digit seconds if it is acceptable to rely on Ryuk to cleanup containers for tests that did not run.

An alternative strategy would be to split GitBucketCoreModuleSpec into separate specs for h2, postgres, and MySQL - that would make targeting better, though such a refactor would still benefit from the start-at-test-registration container strategy here.

### DIFF
--- a/src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
+++ b/src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
@@ -12,17 +12,23 @@ import org.apache.commons.io.FileUtils
 import org.junit.runner.Description
 import org.eclipse.jgit.api.Git
 import org.scalatest.funsuite.AnyFunSuite
-import org.scalatest.Tag
+import org.scalatest.{BeforeAndAfterAll, Tag}
+import org.testcontainers.lifecycle.Startable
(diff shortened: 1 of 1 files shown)


takezoe:
How about reusing docker containers by creating a database per test, like https://github.com/gitbucket/gitbucket/pull/4105? This approach will also allow running tests in parallel.

iamjpotts:
I just compared the MySQL test suite runtime for 5.7+8.4 vs 8.4+9.7. The former runs in around 2:50, the latter runs around 3:10-3:15 - slower by 20-25s. There may be LTS related reasons to drop 5.7 and add 9.7, but it won't help performance of the test suite.

PR #4105, opened 2026-08-11
Reuse docker containers in migration tests

takezoe:

### DIFF
--- a/src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
+++ b/src/test/scala/gitbucket/core/GitBucketCoreModuleSpec.scala
@@ -12,17 +12,19 @@ import org.apache.commons.io.FileUtils
 import org.junit.runner.Description
 import org.eclipse.jgit.api.Git
 import org.scalatest.funsuite.AnyFunSuite
-import org.scalatest.Tag
+import org.scalatest.{BeforeAndAfterAll, Tag}
+import org.testcontainers.lifecycle.Startable
 import org.testcontainers.mysql.MySQLContainer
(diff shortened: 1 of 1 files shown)


iamjpotts:
What's the before/after runtime for the whole test suite with this change?

takezoe:
See https://github.com/gitbucket/gitbucket/actions/runs/31451837984/job/93657614742?pr=4105. It looks like the same level as #4101 and could be faster more by parallelizing.

iamjpotts:
Interesting. This pr (6:55) and mine (6:45) are within 10s of each other.

This one is at 7:29, so either of our prs are substantially better - https://github.com/gitbucket/gitbucket/actions/runs/31324916579/job/93273633803?pr=4100

There was one other idea I was kicking around, I can put up another pr for comparison to our first two, but it would be a few days.

takezoe:
There seems to be some fluctuation on GitHub Actions. Another attempt of this PR was 5m 42s:
https://github.com/gitbucket/gitbucket/actions/runs/31451837984/job/93657614715?pr=4105

iamjpotts:
If the timings are all close, I would lean towards the pr that has the least complexity (easiest to maintain in the future). I want to try the third idea first.

takezoe:
By the way, I started feeling that we may rather consider dropping support for old versions of MySQL and PostgreSQL.

Particularly, MySQL 5.7 has already been EOL and it only has amd64 docker image which is slower on arm machines like Apple Silicon mac. Actually, simply dropping MySQL 5.7 support made `GitBucketCoreModuleSpec` 2x faster on my M2 MacBook Air.

